PR #111, opened 2015-10-06
make rikaikun not to break on not elements

eiennohito:
for example comment nodes inside text break him badly

I wonder if the project is still active

### DIFF
--- a/rikaicontent.js
+++ b/rikaicontent.js
@@ -470,11 +470,15 @@ var rcxContent = {
 		'RB': true,
 		'RT': true,
 		'RP': true
-	},
-
+	},	
 	isInline: function(node) {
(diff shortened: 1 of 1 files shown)


eiennohito:
added comment, squashed commits

melink14:
Do you have some example urls that should break so I can test them from head for a bit before pushing this update.

eiennohito:
Actually, any Angular.js thing will do.

Angular.js inserts comments in dom tree.

http://lotus.kuee.kyoto-u.ac.jp/depfinder/search?q=%E3%80%9C%E7%A7%81%E3%81%8C%E2%86%92%EF%BC%8A%E5%AF%9D%E3%82%8B <http://lotus.kuee.kyoto-u.ac.jp/depfinder/search?q=%E3%80%9C%E7%A7%81%E3%81%8C%E2%86%92%EF%BC%8A%E5%AF%9D%E3%82%8B>

PR #117, opened 2016-07-03
Option for showing popup only when key is pressed and configurable delay to open popup.

versusvoid:
Long awaited feature and annoyment reduction.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,7 +2,7 @@
 	"manifest_version": 2,
 	
 	"name": "rikaikun",
-	"version": "0.8.91",
+	"version": "0.8.92",
 	
 	"description": "A port of rikaichan for chrome. Translate Japanese by hovering over words.",
 	"icons": { "48": "images/icon48.png",
--- a/options.html
+++ b/options.html
@@ -39,6 +39,12 @@
 					<!-- <div id="keycon" class="tabcontent hide"> -->
 					<div id="keycon" class="tabContent">
 					<h1>Keyboard</h1>
+						Show popup only on pressed key:<br />
+						<input type="radio" name="showOnKey" value="">Not used<br />
+						<input type="radio" name="showOnKey" value="Alt">Alt<br />
+						<input type="radio" name="showOnKey" value="Ctrl">Ctrl<br />
+						<input type="radio" name="showOnKey" value="Alt+Ctrl">Alt+Ctrl<br />
+						<br />
 						<strong>Keys when popup is visible</strong><br><br>
 						<table cellspacing=5>
 						<tr><td>A</td><td>Alternate popup location</td></tr>
--- a/options.html
+++ b/options.html
@@ -30,6 +30,10 @@
 							<br />
 							<input type="checkbox" id="minihelp" name="minihelp">
 							Show mini help
+							<br />
+							<br />
+							Show popup with delay:
+							<input type="number" id="popupDelay" name="popupDelay" min="1"> milliseconds
 							</p>
 					</div>
 					<!-- <div id="keycon" class="tabcontent hide"> -->
--- a/options.js
+++ b/options.js
@@ -27,6 +27,8 @@ function fillVals() {
 	else
 		document.optform.minihelp.checked = false;
 
+	document.optform.popupDelay.value = localStorage["popupDelay"];
+
 	if (localStorage['disablekeys'] == 'true')
 		document.optform.disablekeys.checked = true;
 	else
@@ -83,12 +85,24 @@ function getVals() {
 	localStorage['copySeparator'] = document.optform.copySeparator.value;
 	localStorage['maxClipCopyEntries'] = document.optform.maxClipCopyEntries.value;
 
+	var popupDelay;
+	try {
+		popupDelay = parseInt(document.optform.popupDelay.value);
+		if (!isFinite(popupDelay)) {
+			throw Error('infinite');
+		}
+		localStorage['popupDelay'] = document.optform.popupDelay.value;
+	} catch (err) {
+		popupDelay = 150;
+		localStorage['popupDelay'] = "150";
+	}
 
 	chrome.extension.getBackgroundPage().rcxMain.config.css = localStorage["popupcolor"];
 	chrome.extension.getBackgroundPage().rcxMain.config.highlight = localStorage["highlight"];
 	chrome.extension.getBackgroundPage().rcxMain.config.textboxhl = localStorage["textboxhl"];
 	chrome.extension.getBackgroundPage().rcxMain.config.onlyreading = localStorage["onlyreading"];
 	chrome.extension.getBackgroundPage().rcxMain.config.minihelp = localStorage["minihelp"];
+	chrome.extension.getBackgroundPage().rcxMain.config.popupDelay = popupDelay;
 	chrome.extension.getBackgroundPage().rcxMain.config.disablekeys = localStorage["disablekeys"];
 	chrome.extension.getBackgroundPage().rcxMain.config.kanjicomponents = localStorage["kanjicomponents"];
 	chrome.extension.getBackgroundPage().rcxMain.config.kanjiinfo = kanjiinfoarray;
--- a/options.js
+++ b/options.js
@@ -63,6 +63,14 @@ function fillVals() {
 
 	document.optform.maxClipCopyEntries.value = parseInt(localStorage['maxClipCopyEntries']);
 
+	store = localStorage['showOnKey'];
+	for(var i = 0; i < document.optform.showOnKey.length; ++i) {
+		if (document.optform.showOnKey[i].value === store) {
+			document.optform.showOnKey[i].checked = true;
+			break;
+		}
+	}
+
 }
 
 function getVals() {
@@ -96,6 +104,7 @@ function getVals() {
 		popupDelay = 150;
 		localStorage['popupDelay'] = "150";
 	}
+	localStorage['showOnKey'] = document.optform.showOnKey.value;
 
 	chrome.extension.getBackgroundPage().rcxMain.config.css = localStorage["popupcolor"];
 	chrome.extension.getBackgroundPage().rcxMain.config.highlight = localStorage["highlight"];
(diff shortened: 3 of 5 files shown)


melink14:
As a high level comment:
It'd be better if we kept separate features (even if related) in separate pull requests (or perhaps at least separate commits?).  This allows it to be very easy to determine which changes are responsible for which features.

manifest.json
```diff
@@ -2,7 +2,7 @@
 	"manifest_version": 2,
 	
 	"name": "rikaikun",
-	"version": "0.8.91",
+	"version": "0.8.92",
```
melink14: Updating this in a PR doesn't really make sense because depending on which features are collected in a release this number will change before publishing.

That being said it's not a big deal because you need some new number for your new options anyway; and it's easy to fix up later before publishing.

options.html
```diff
@@ -30,11 +30,25 @@
...
+							<br />
+							Show popup with delay: 
+							<input type="number" id="popupDelay" name="popupDelay" min="1"> milliseconds
 							</p>
 					</div>
 					<!-- <div id="keycon" class="tabcontent hide"> -->
 					<div id="keycon" class="tabContent">
 					<h1>Keyboard</h1>
+						<input type="checkbox" id="showOnKeyFlag" name="showOnKeyFlag" disabled>
+						Immediately show only when key <select id="showOnKey" name="showOnKey">
```
melink14: This isn't just 'Immediately' right?  Isn't the goal to show definitions *only* when the key is pressed?

options.html
```diff
@@ -30,11 +30,25 @@
...
+							<br />
+							<br />
+							Show popup with delay: 
+							<input type="number" id="popupDelay" name="popupDelay" min="1"> milliseconds
 							</p>
 					</div>
 					<!-- <div id="keycon" class="tabcontent hide"> -->
 					<div id="keycon" class="tabContent">
 					<h1>Keyboard</h1>
+						<input type="checkbox" id="showOnKeyFlag" name="showOnKeyFlag" disabled>
```
melink14: This design seems like it could be simplified.

Instead of a checkbox followed by a select why not use radio buttons where the default value is  "Not used" or similar?

options.html
```diff
@@ -30,11 +30,25 @@
...
+						Immediately show only when key <select id="showOnKey" name="showOnKey"> 
+							<option value="">--</option> 
+							<option value="Alt">Alt</option> 
+							<option value="Ctrl">Ctrl</option> 
+							<option value="Alt+Ctrl">Alt+Ctrl</option> 
+						</select> is pressed.
+
+						<br />
```
melink14: prefer not closing your <br> tags.

options.js
```diff
@@ -82,14 +96,24 @@ function getVals() {
 	localStorage['lineEnding'] = document.optform.lineEnding.value;
 	localStorage['copySeparator'] = document.optform.copySeparator.value;
 	localStorage['maxClipCopyEntries'] = document.optform.maxClipCopyEntries.value;
-
+	try {
+		if (!isFinite(parseInt(document.optform.popupDelay.value))) {
```
melink14: why not save the parsed popupDelay value and then add it to localStorage.  That way you don't need to parseInt twice.

rikaicontent.js
```diff
@@ -236,7 +236,12 @@ var rcxContent = {
 				// go up if necessary
 				if ((y + v + pH) > window.innerHeight) {
 					var t = y - pH - 30;
-					if (t >= 0) y = t;
+					if (t >= 0) {
```
melink14: how does this section relate to popup delays and keyboard shortcuts?

rikaicontent.js
```diff
@@ -285,22 +290,22 @@ var rcxContent = {
 			// The reason for this is becasue if there's no selection 
 			// we probably clicked somewhere else and we don't want to 
 			// bounce back.
-		    if(!sel.toString())
```
melink14: I appreciate formatting changes but they should be in a separate request.

rikaicontent.js
```diff
@@ -314,12 +319,28 @@ var rcxContent = {
...
+	lastTarget: null,
 
 	onKeyDown: function(ev) { rcxContent._onKeyDown(ev) },
 	_onKeyDown: function(ev) {
-//		this.status("keyCode=" + ev.keyCode + ' charCode=' + ev.charCode + ' detail=' + ev.detail);
-
-		if ((ev.altKey) || (ev.metaKey) || (ev.ctrlKey)) return;
+		//console.log("key=" + ev.key + " keyCode=" + ev.keyCode + ' charCode=' + ev.charCode + ' detail=' + ev.detail);
+ 
+		if (window.rikaichan.config.showOnKey !== "" && (ev.altKey || ev.ctrlKey)) {
```
melink14: instead of checking if showOnKey is "", this seems like it would be better to store the keycode of the whichever key was chosen in options right?

rikaicontent.js
```diff
@@ -788,29 +809,46 @@ var rcxContent = {
...
+
+		if ((window.rikaichan.config.showOnKey.includes("Alt") && !ev.altKey) ||
+			 (window.rikaichan.config.showOnKey.includes("Ctrl") && !ev.ctrlKey)) {
+			this.clearHi();
+			this.hidePopup();
+			return;
+		}
+
 		var fake;
+		var tdata = window.rikaichan;	// per-tab data
```
melink14: why did you move these vars out?

rikaicontent.js
```diff
@@ -976,9 +1003,13 @@ var rcxContent = {
 			tdata.popX = ev.clientX;
 			tdata.popY = ev.clientY;
 			tdata.timer = setTimeout(
-				function(tdata) {
+				function(tdata, title) {
+					if (!window.rikaichan || title !== window.rikaichan.title) {
+						//console.log("Title changed during delay");
```
melink14: remove comment?

rikaicontent.js
```diff
@@ -933,25 +957,28 @@ var rcxContent = {
...
 		}  */
-
 		tdata.prevTarget = ev.target;
 		tdata.prevRangeNode = rp;
 		tdata.prevRangeOfs = ro;
 		tdata.title = null;
 		tdata.uofs = 0;
 		this.uofsNext = 1;
 
+		var delay = window.rikaichan.config.showOnKey === "" ? window.rikaichan.config.popupDelay : 1;
```
melink14: why not allow users to set a pop up delay even if they use a showOnKey?

rikaicontent.js
```diff
@@ -933,25 +957,28 @@ var rcxContent = {
...
 			this.forceKanji = ev.shiftKey ? 1 : 0;
 			tdata.popX = ev.clientX;
 			tdata.popY = ev.clientY;
 			tdata.timer = setTimeout(
-				function() {
-					//chrome.extension.sendMessage({"type":"resetDict"});
-					//rcxContent.show(tdata, this.forceKanji ? this.forceKanji : this.defaultDict);
+				function(rangeNode, rangeOffset) {
+					if (!window.rikaichan || rangeNode != window.rikaichan.prevRangeNode || ro != window.rikaichan.prevRangeOfs) {
+						//console.log("Range changed during delay");
```
melink14: maybe make this console.log a regular implementation comment for this block?

rikaicontent.js
```diff
@@ -321,19 +321,23 @@ var rcxContent = {
 	keysDown: [],
 	lastPos: { x: null, y: null},
 	lastTarget: null,
+	altGraphKey: false,
 
 	onKeyDown: function(ev) { rcxContent._onKeyDown(ev) },
 	_onKeyDown: function(ev) {
 		//console.log("key=" + ev.key + " keyCode=" + ev.keyCode + ' charCode=' + ev.charCode + ' detail=' + ev.detail);
  
+		if (ev.key === "AltGraph") {
```
melink14: Should AltGr be treated as a separate key in the options or is it normal to treat it as identical to alt for these types of things?

rikaicontent.js
```diff
@@ -321,19 +321,23 @@ var rcxContent = {
...
+			this.altGraphKey = true;
+		}
 		if (window.rikaichan.config.showOnKey !== "" && (ev.altKey || ev.ctrlKey)) {
 			if (this.lastTarget !== null) {
 				//console.log(ev);
 				var myEv = {
 					clientX: this.lastPos.x,
 					clientY: this.lastPos.y,
 					target: this.lastTarget,
-					altKey: ev.altKey || ev.key == "Alt",
```
melink14: If we want altGraph to be the same as alt here why don't we do something like: ev.altKey || ev.modifierAltGraph

That should work if I'm reading the spec correctly.

I'm not sure we need to add a special variable.

options.html
versusvoid: Yes, but without delay, which now defaults to 150ms.

rikaicontent.js
versusvoid: `tdata` is also used outside `try {}` scope and I have no idea why JS allows such code, but this is definitely wrong.

rikaicontent.js
versusvoid: Identical. This is chromium's stupidity. On russian layout it generates `Alt` event for left Alt and `AltGraph` for right Alt. Even 'Right Alt+Left' doesn't take you back in history on russian layout, while 'Left Alt+Left' does.

rikaicontent.js
versusvoid: I'll check if chromium supports it already.

rikaicontent.js
melink14: Because JS is amazing of course, anything is possible.  (github diffs are hard for me to read so at first I actually didn't see it was outside the try)

This code could really benefit from a re-write since I barely refactored at all when I ported rikaichan.

options.html
melink14: I see.  Perhaps we should explicitly mention that it overrides the popup delay setting; at least for me I didn't make the connection.

Also, it seems like no delay is correct for the initial press of the modifier key, but if the user is holding down the modifier key while scanning a sentence, won't they want the same delay as a normal user?

I assume you played with this during implementation and it felt better?

versusvoid: My reasoning was: user has to make _some minimal action_ to show popup. For usual mode I hover mouse and hold it over a word for at least 150ms. But in other mode I'm already doing _some action_ (keeping button pressed), so I expect app response immediately.

But it doesn't matter anymore, I'll make options separate.

options.html
versusvoid: Why, by the way? 
xkcd

versusvoid:
@melink14 have fixed everything what can be fixed.

options.html
melink14: I guess the core is that it's an extra two characters that aren't needed.

A larger reason is that it's an XHTML idiom that feels out of place in pure
HTML.

I'm sure the requirements of my day job also color my review as well.

It's not really fair to hold everyone to some standard without even a
published style guide. :)

2016年9月9日(金) 12:57 Versus <notifications@github.com>:

PR #121, opened 2016-07-29
Async dictionary loading

JakeH:
Hi there. 

I've noticed that rikaikun takes a bit of time to become responsive after enabling. On my machine, it would take an average of 1500ms to completely load all dictionaries. 

To help reduce this, I've updated the dictionary initialization to use async calls and Promises to reduce the load time to around 800ms. 

Please review and let me know what you think.

### DIFF
--- a/data.js
+++ b/data.js
@@ -40,19 +40,62 @@
 */
 
 function rcxDict(loadNames) {
-	this.loadDictionary();
-	if (loadNames) this.loadNames();
-	this.loadDIF();
+
+	
 }
 
 rcxDict.prototype = {
 	config: {},
 
-	setConfig: function(c) {
+	init: function (loadNames) {
+		var started = +new Date();
+
+		var promises = [
+			this.loadDictionary(loadNames),
+			this.loadDIF()
+		];
+
+		return Promise.all(promises)
+			.then(function () {
+				var ended = +new Date();
+				console.log('rcxDict main then in ' + (ended - started));
+			}.bind(this));
+	},
+
+	setConfig: function (c) {
 		this.config = c;
 	},
 
 	//
+
+	fileReadAsync: function (url, asArray) {
+		return new Promise(function prom(resolve, reject) {
+			var req = new XMLHttpRequest();
+
+			req.onreadystatechange = function () {
+
+				if (req.readyState === 4) {
+
+					if (!asArray)
+						resolve(req.responseText);
+					else {
+						var a = req.responseText.split('\n')
+							.filter(function removeBlanks(o) {
+								return o && o.length > 0
+							});
+
+						resolve(a);
+					}
+
+
+				}
+			};
+
+			req.open("GET", url, true);
+			req.send(null);
+
+		});
+	},
 
 	fileRead: function(url, charset) {
 		var req = new XMLHttpRequest();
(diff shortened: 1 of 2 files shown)


melink14:
Thanks.  I'm traveling this weekend but I'll try to review and process all
pull requests in the next week at least.

Your contributions are appreciated.

2016年7月29日(金) 1:24 Jake <notifications@github.com>:

data.js
```diff
@@ -40,20 +40,63 @@
...
+	init: function (loadNames) {
+		var started = +new Date();
+
+		var promises = [
+			this.loadDictionary(loadNames),
+			this.loadDIF()
+		];
+
+		return Promise.all(promises)
+			.then(function () {
```
melink14: consider arrow functions for over anon function + bind

data.js
```diff
@@ -40,20 +40,63 @@
...
+				}
+			};
+
+			req.open("GET", url, true);
+			req.send(null);
+
+		});
+	},
+
 	fileRead: function(url, charset) {
```
melink14: Is this unused now?

data.js
```diff
@@ -40,20 +40,63 @@
...
 
+	fileReadAsync: function (url, asArray) {
+		return new Promise(function prom(resolve, reject) {
+			var req = new XMLHttpRequest();
+
+			req.onreadystatechange = function () {
+
+				if (req.readyState === 4) {
+
+					if (!asArray)
```
melink14: it might be helpful for future editors if we use braces for ifs that precede elses.

data.js
```diff
@@ -40,20 +40,63 @@
...
+						resolve(req.responseText);
+					else {
+						var a = req.responseText.split('\n')
```
melink14: Why is the variable called a?

data.js
```diff
@@ -99,14 +142,41 @@ rcxDict.prototype = {
 		this.nameIndex = this.fileRead(chrome.extension.getURL("data/names.idx"));
 	},
 
+	loadFileShortcut: function (file, isArray, target) {
```
melink14: what does the 'Shortcut' suffix denote?

data.js
```diff
@@ -99,14 +142,41 @@ rcxDict.prototype = {
...
+		var url = chrome.extension.getURL('data/' + file);
+
+		return this.fileReadAsync.call(this, url, isArray).then(function (data) {
```
melink14: why use 'call' instead of regular method invocation?

data.js
```diff
@@ -198,33 +268,36 @@ if (0) {
...
-				o.to = f[1];
-				o.type = f[2];
-				o.reason = f[3];
-
-				if (prevLen != o.from.length) {
-					prevLen = o.from.length;
-					g = [];
-					g.flen = prevLen;
-					this.difRules.push(g);
+		return this.fileReadAsync(chrome.extension.getURL("data/deinflect.dat"), true)
```
melink14: /* isArray */ before the true if it's not too much trouble.

rikaichan.js
```diff
@@ -308,22 +308,30 @@ var rcxMain = {
 	inlineEnable: function(tab, mode) {
 		if (!this.dict) {
 			//	var time = (new Date()).getTime();
-			if (!this.loadDictionary()) return;
+			var loadProm = this.loadDictionary();
+
+			if (loadProm == null)
```
melink14: It'd be better to have return on the same line or add braces.  it's very easy to add another line to an uncontained if without realizing it doesn't work.

rikaichan.js
```diff
@@ -164,15 +164,15 @@ var rcxMain = {
 				return false;
 			} */
 			try {
-				this.dict = new rcxDict(this.haveNames/* && !this.cfg.nadelay*/);
+				this.dict = new rcxDict();
+				return this.dict.init(this.haveNames/* && !this.cfg.nadelay*/);
 				//this.dict.setConfig(this.dconfig);
 			}
 			catch (ex) {
 				alert('Error loading dictionary: ' + ex);
```
melink14: This is no longer reachable right?

It seems like instead we'd want to use thenCatch at the call site?

rikaichan.js
```diff
@@ -164,15 +164,15 @@ var rcxMain = {
...
-				return false;
 			}
 		}
-		return true;
+		return null;
```
melink14: I think instead of returning null here we should pass true/false.

something like this above:

return this.dict.init(blahblah).then(() => {return false}, () => {alert('Error loading dictionary: ' + ex);})

and then here: return:
Promise.resolve(true);

melink14:
As a high level comment:

Thanks a lot for doing this!  I tend to write too much in comments but the commit is basically okay.  If you agree with my comments you can make the change or discuss it or anything else.

After all, the code base didn't start that great to begin with! :)

PR #124, opened 2016-09-06
Add scroll support for many definitions

bazzinotti:
# Scenario

A rikaikun user hovers over a japanese phrase. If a given phrase has more than 7 definitions, only the first 7 can be viewed, and all of the others are unable to be viewed.

This PR allows the definitions window to be scrollable forwards and backwards when there are too many definitions to be displayed. This allows all definitions to be viewed.

Rikaikun uses the ellipsis '...' to inform the user there are more definitions than can be displayed. When this is the case, the 'J' and 'K' keys can be used to scroll through the definitions forwards and backwards respectively (with loopback both ways).
## Technical Details
- ellipsis on head and tail of definition window helps inform the user where they are. (first, middle, last)
### Todo

When I approached this codebase, the max number of display-able definitions was hard-coded at 7. Unfortunately, I have continued this practice. I assume you understand why I'd like to remove these "floating" numbers and box them into a single variable. However, I am unsure of where to add the proper constant variable declaration. Please inform me of a suitable location and I will add the commit.

Please feel free to chime in any additional notes! Thanks!

### DIFF
--- a/data.js
+++ b/data.js
@@ -428,7 +428,6 @@ if (0) {
                     if (ok) {
                         if (count >= maxTrim) {
 							entry.more = 1;
-							//break;
 						}
 
 						have[ofs] = 1;
--- a/data.js
+++ b/data.js
@@ -428,7 +428,7 @@ if (0) {
                     if (ok) {
                         if (count >= maxTrim) {
 							entry.more = 1;
-							break;
+							//break;
 						}
 
 						have[ofs] = 1;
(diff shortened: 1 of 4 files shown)


melink14:
Perhaps make that size configurable in the options so that people can choose how many entries they want and it gives a global place to define it. I guess you could put it in the `config` object even without making it an option.

melink14:
Merged from head and forked PR to https://github.com/melink14/rikaikun/pull/162 for further review.

PR #126, opened 2017-05-04
Added TTS support using responsivevoice.org

MayamaTakeshi:
Hello,
I have modified rikaikun to make use of TTS provided by
  https://responsivevoice.org/license/
which is free for non-commercial projects.
As required by the license, rikaikun will add a link to responsivevoice.org whenever TTS is enabled.

### DIFF
--- a/background.html
+++ b/background.html
@@ -1,5 +1,6 @@
 <html>
 	<head>
+		<script src='https://code.responsivevoice.org/responsivevoice.js'></script>
 		<script src="data.js"></script>
 		<script src="rikaichan.js"></script>
 		<script src="background.js"></script>
--- a/background.js
+++ b/background.js
@@ -1,3 +1,5 @@
+var last_time = new Date().valueOf();
+var previous_text = null;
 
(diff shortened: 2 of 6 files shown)


MayamaTakeshi:
Hello,
I got some clarification from ResponsiveVoice and it cannot be used for free in browser extensions:

> Hi Takeshi,
>
> There's a correction from the development team, they mistakenly referred to an older payment policy. > At this time, ALL browser extensions are required to 1) Provide a fixed referrer, and 2) have an active ResponsiveVoice Pro subscription.
>
> Apologies for the confusion!
>
> Kind regards,
> Suzanne
ResponsiveVoice Support


I was hoping to use ResponsiveVoice because it uses local TTS engine voice if available and if not, it  falls back to server-generated TTS. So the same code could be used for firefox (rikaichan)  or in case of chrome installation without Japanese TTS support (availability of voices might be restricted by region or other factors).
So, I am closing this PR and will open another to just use local TTS from chrome.
Sorry for the noise.

PR #127, opened 2017-05-10
Added support for TTS on hover

MayamaTakeshi:
Hello,
here are my changes to permit to play native TTS on hover. It will work as long as the chrome installation has access to TTS lang ja-JP.


background.js
```diff
@@ -1,3 +1,5 @@
+var last_time = new Date().valueOf();
```
melink14: It seems weird to have these as global variables...

Ideally we'd encapsulate the TTS support but at least we should move them to be next to the function that needs them for now.

rikaicontent.js
```diff
@@ -748,6 +748,14 @@ var rcxContent = {
 		sel.removeAllRanges();
 		sel.addRange(range);
 		tdata.selText = sel.toString();
+
+		if(window.rikaichan.config.ttsEnabled == 'true') {
+			var text = sel.toString();
+			if(typeof text === 'string' && text.length > 0) {
+				console.log("Sending playTTS " + text);
```
melink14: it turns out it's bad practice to log in content scripts, though useful for debugging.

background.js
```diff
@@ -108,6 +121,23 @@ function initStorage(key, initialValue) {
   return false;
 }
 
+function playTTS(text) {
+	var now = new Date().valueOf();
```
melink14: I wonder if we should start using const/let in new code...

background.js
MayamaTakeshi: OK. I will move the code to a module.

rikaicontent.js
MayamaTakeshi: OK. I will comment that line.

MayamaTakeshi:
> 1. Does it get weird for it to always speak? Should it be a keyboard shortcut?

Well, an option was added for configuration. By default Text-To-Speech is disabled.

> What happens if the user doesn't have JA TTS?

It just doesn't play anything (I tested setting language to fake "mo-MO" instead of "ja-JP") but doesn't interfere with the other functionality.
Probably we should check and inform this when the user tries to enable TTS on the options page but let me leave this for another PR.

rikaicontent.js
```diff
@@ -750,6 +750,14 @@ var rcxContent = {
 		sel.removeAllRanges();
 		sel.addRange(range);
 		tdata.selText = sel.toString();
+
+		if(window.rikaichan.config.ttsEnabled == 'true') {
+			var text = sel.toString();
+			if(typeof text === 'string' && text.length > 0) {
```
melink14: is it possible for text to not be string or is it just that text may be undefined?

rikaicontent.js
```diff
@@ -750,6 +750,14 @@ var rcxContent = {
 		sel.removeAllRanges();
 		sel.addRange(range);
 		tdata.selText = sel.toString();
+
+		if(window.rikaichan.config.ttsEnabled == 'true') {
+			var text = sel.toString();
+			if(typeof text === 'string' && text.length > 0) {
+				//console.log("Sending playTTS " + text);
```
melink14: No need for this comment, I guess.

texttospeech.js
```diff
@@ -0,0 +1,25 @@
...
+			console.log("tts.speak(" + text + ")");
+			window.speechSynthesis.cancel();
+			let u = new SpeechSynthesisUtterance();
+			u.text = text;
+			u.lang = 'ja-JP';
+			window.speechSynthesis.speak(u);
+			this.previous_text = text;
+			this.last_time = now;
+		} else {
+			console.log("Ignoring " + text);
```
melink14: no need to do now, but maybe it'd be good to pop up a message if we know the reason why it didn't work (no tts library etc)

rikaicontent.js
```diff
@@ -750,6 +750,14 @@ var rcxContent = {
 		sel.removeAllRanges();
 		sel.addRange(range);
 		tdata.selText = sel.toString();
+
+		if(window.rikaichan.config.ttsEnabled == 'true') {
```
melink14: is it annoying if it talks all the time? i wonder if it should be a shortcut or button. I'll test it out as well.

melink14: I guess we mentioned this before but I just meant that even if it was enabled maybe you don't want every single word spoken. I'll test it though.

texttospeech.js
melink14: (this was also mentioned before and I agree)

rikaicontent.js
melink14: Sounded pretty good. I'm convinced.

rikaicontent.js
melink14: Well, I guess the idea is it will be easy to debug later. probably we should add a debug mode which would allow us to only send logs without adding commented lines which is a code smell at work.

Shouldn't be so strict here though.

rikaicontent.js
melink14: I couldn't find a time this failed and https://developer.mozilla.org/en-US/docs/Web/API/Selection/toString says that it must be a string so let's remove.

melink14:
I'm going to resolve the final comment and merge from HEAD for now.

It'd be good for the options.js refactor to happen first so may try to merge this tomorrow.

background.js
melink14: have other issues open to modernize code base.

background.js
```diff
@@ -88,6 +91,13 @@ if(initStorage("v0.8.92", true)) {
     // v0.8.92
 	initStorage("popupDelay", "150");
 	initStorage("showOnKey", "");
+
+
+	// v???
+	initStorage("ttsEnabled", "false");
+	if (localStorage['ttsEnabled'] == "yes") {
```
melink14: Just fyi, that you don't need this `== "yes"` part for new options. It was only used when I changed from using "yes" to "true".

melink14: (I fixed it already though)

PR #134, opened 2017-10-18
Added Heisig keywords to the numbers. Added 6th edition Heisig option.

Vwing:
Hello,

This is a feature I've been dying to have for quite some time. Decided to implement it myself. It's a huge time-saver to have the Heisig keyword on Rikaikun instead of having to search for the number on kanji koohii or somesuch.

image

Let me know what you all think.


Vwing:
Just pushed a commit with the rest of the keywords filled in (2043 onward). Also renamed "Heisig" to "Heisig 5th" in preparation for when I add all the Heisig 6th edition keywords and numbers.

Vwing:
Okay, 6th edition keywords and numbers are in.

I named the entry.misc / numList option 'LF' because 'F' is the 6th letter in the alphabet. I was going to rename the Heisig 5th edition from 'L' to 'LE' , but somehow that broke things...

christopherball:
I came across this code review during my searches as I realized the rikaikun chrome plugin is providing incorrect heisig numbers for owners of the v6 edition.  Was there a reason this feature was never merged in?  Would be great to have this feature (and correct 6th edition numbers!).

melink14:
The reason is simple, there are unresolved comments from my code review on March 3rd that I need answered before I can merge it in.

For small stuff, I don't mind merging and doing fixes myself since I appreciate any contributions to make rikaikun better, but for my comment above I don't want to merge in manual changes to generated data without a plan forward since they'll be wiped out next time the data is regenerated.

melink14:
To answer my own question, the data came from https://github.com/sdcr/heisig-kanjis/blob/master/heisig-kanjis.csv which would be easy to use to modify these keywords.

Thanks.

PR #147, opened 2018-07-06
Fix #140 - Allow default copying behaviour when pressing ctrl + c

### LINKED ISSUE #140: 'c' key should only copy if ctrl is also not pressed
It makes it hard to copy Japanese text without triggering the popup copy shortcut.

darren-lester:


rikaicontent.js
```diff
@@ -405,6 +410,10 @@ var rcxContent = {
 
 		this.keysDown[ev.keyCode] = 1;
 
+		if (!shouldPreventDefault) {
+			return;
+		}
+
 		// don't eat shift if in this mode
 		if (true/*!this.cfg.nopopkeys*/) {
```
melink14: Should we just roll your new variable into this if statement since it seems like it was meant to be conditional anyway?

darren-lester: That sounds good to me. I had avoided modifying the if statement on line 418 as I did not understand the nature of the comment and so decided to leave it be.

PR #151, opened 2018-11-17
Fix #135 - Migrate to options version 2

### LINKED ISSUE #135: Migrate to options version 2
https://developer.chrome.com/extensions/optionsV2

darren-lester:


melink14:
(I thought I replied to this a long time ago)

#135 was mostly thinking about cloud sync which #161 is meant to do though this PR switches to the popup options view which may also be good; I'll test it out and merge if it seems better.

melink14:
I like that the user stays in context though it is a bit small causing more scrolling perhaps. I think it's probably worth merging.

PR #161, opened 2019-09-15
Initial commit for options upgrade to cloud storage

deshaun93:


background.js
```diff
@@ -1,128 +1,132 @@
...
-rcxMain.config.kanjicomponents = localStorage["kanjicomponents"];
-rcxMain.config.kanjiinfo = new Array(rcxDict.prototype.numList.length/2);
-for (i = 0; i*2 < rcxDict.prototype.numList.length; i++) {
-	rcxMain.config.kanjiinfo[i] = localStorage[rcxDict.prototype.numList[i*2]];
-}
-
+chrome.browserAction.onClicked.addListener(rcxMain.inlineToggle);
+chrome.tabs.onSelectionChanged.addListener(rcxMain.onTabSelect);
+chrome.runtime.onMessage.addListener(
+    function (request, sender, response) {
```
melink14: oh no!

You changed all the tabs into spaces which makes every single line a diff...

The main goal was to get your intermediate changes on github so I'll merge this into a different branch so that we can iterate on it.

melink14: Ah, actually this PR is already a branch; let me know if you can fix the spacing easily or if you mind if I push changes to it.

deshaun93: Taking a look now.

background.js
```diff
@@ -46,83 +45,88 @@ chrome.runtime.onMessage.addListener(
...
-	// v0.7
-	initStorage("popupcolor", "blue");
-	initStorage("highlight", true);
-	
-	// v0.8
-	// No changes to options
-	
-	// V0.8.5
-	initStorage("textboxhl", false);
+	})
```
melink14: why did you remove this semicolon?

background.js
```diff
@@ -46,83 +45,88 @@ chrome.runtime.onMessage.addListener(
...
+var getItems = ["popupcolor", "highlight", "textboxhl", "onlyreading",
```
melink14: Two things:

1) We should share this array of option with options.js using a global object like `optionNames`
2) We should probably make it easier to read by initializing it with one option name per line in alphabetical order.

background.js
```diff
@@ -46,83 +45,88 @@ chrome.runtime.onMessage.addListener(
...
+			// V0.8.5
+			initStorage("textboxhl", false);
 
-	for (i = 0; i*2 < rcxDict.prototype.numList.length; i++) {
-		initStorage(rcxDict.prototype.numList[i*2], "true")
-	}
+			// v0.8.6
+			initStorage("onlyreading", false);
+			// v0.8.8
+			if (items.highlight == "yes")
```
melink14: These lines don't make sense. You're reading a value like "yes" from storage and then you're rewriting the same local object to have a different but similar value.

I know you were modeling it after what was happening to localStorage before but there are 2 main differences:
1) localStorage is a global variable so updating it's value actually changes real state instead of just local state.
2) localStorage was updated like that because I changed the names a long time ago and it was required for backwards compatibility. We don't need to do that for a brand new storage layer except when interacting with legacy storage.

So in this case, we don't need these new if statements but we may want the old ones for backwards compatibility, or we can argue that it's been long enough and drop support for the legacy yes no comments.

background.js
```diff
@@ -46,83 +45,88 @@ chrome.runtime.onMessage.addListener(
...
-	rcxMain.config.kanjiinfo[i] = localStorage[rcxDict.prototype.numList[i*2]];
-}
+		/**
+		 * Initializes the localStorage for the given key.
+		 * If the given key is already initialized, nothing happens.
+		 *
+		 * @author Teo (GD API Guru)
+		 * @param key The key for which to initialize
+		 * @param initialValue Initial value of localStorage on the given key
+		 * @return boolean if a value is assigned or false if nothing happens
```
melink14: revert this line (though you can keep the indent)

It doesn't make sense to say `@return boolean` instead of `@return true`.

background.js
```diff
@@ -46,83 +45,88 @@ chrome.runtime.onMessage.addListener(
...
+		 */
+		function initStorage(key, initialValue) {
```
melink14: You've moved initStorage inside of this callback which is a bit weird. I guess it's only used inside the callback but it does look a bit weird to have the function defined here. I would move it to global scope for consistency.

melink14: This function should probably be changed to work with cloud sync. (Right now, you read from cloud sync but you don't handle updating it with default values; if we change this, we can have it work with legacy localstorage values and default values)

It can't work exactly the same since you can't write to cloud storage one field at a time, so we'll need to accumulate the set of initial values and then save them back to cloud storage (either always, or only if the final values are different than what we read from the cloud).

The way I would do that would be to reuse `rcxMain.config` since it's the local store for settings anyway. You can initialize it earlier and then update it one field at a time inside of `initStorage`. Then, after all the settings are initialized we can save the settings back to cloud storage as mentioned before.

The high level goal would be to use cloud setting if it exist, then check localStorage, and then just use default value.

options.js
```diff
@@ -1,139 +1,193 @@
...
-	// change format of selected tab to selected
-	// hide all tab contents
-	// show selected tab contents
-}*/
-
-
-document.querySelector('#submit').addEventListener('click', getVals);
-
+
+//Todo: Update config
```
melink14: I think you did some of the TODO's but in general, double check and remove them if they're done.

options.js
```diff
@@ -1,139 +1,193 @@
...
+//TOdo: add the second tobi addition
+// Get the current options.
+function get_options() {
```
melink14: JS style is camel case like `getOptions` but in this case since this doesn't return a value maybe it would be better like:

`initializeFormFromCloudSync` or similar.

options.js
```diff
@@ -1,139 +1,193 @@
...
+		function(items) {
+
+			for(var i=0; i < document.optform.popupcolor.length; ++i) {
+				if(document.optform.popupcolor[i].value == items['popupcolor']) {
+					document.optform.popupcolor[i].selected = true;
+					break;
+				}
+			}
+
+			document.optform.popupLocation.selectedIndex = items.popupLocation;
```
melink14: nice work, so much better than the weird string based if checks from before.

options.js
```diff
@@ -1,139 +1,193 @@
...
+			document.optform.popupDelay.value = items.popupDelay;
+
+
+			kanjiInfoLabelList = chrome.extension.getBackgroundPage().rcxDict.prototype.kanjiInfoLabelList;
+
+			for (i = 0; i*2 < kanjiInfoLabelList.length; i++) {
+				// Need to get every other element in the storage, so this funky math was added.
+				// We have abbreviations and values..
+
+				document.getElementById(kanjiInfoLabelList[i*2]).checked = items.kanjiInfo[kanjiInfoLabelList[i*2]] //== 'true' ? true : false;
```
melink14: don't think you need comment at the end.

options.js
```diff
@@ -1,139 +1,193 @@
...
+			}
+
+	for(var i=0; i < document.optform.lineEnding.length; ++i) {
+		if(document.optform.lineEnding[i].value === items.lineEnding) {
```
melink14: this for loop and if statement don't seem to be terminated properly.

the save_options function seems to appear in the middle?

options.js
```diff
@@ -1,139 +1,193 @@
...
+				chrome.extension.getBackgroundPage().rcxMain.config.lineEnding = lineEnding;
+				chrome.extension.getBackgroundPage().rcxMain.config.copySeparator = copySeparator;
+				chrome.extension.getBackgroundPage().rcxMain.config.maxClipCopyEntries = maxClipCopyEntries;
+				chrome.extension.getBackgroundPage().rcxMain.config.popupLocation = popupLocation;
+
+			}
+
+
+			function getPopUpDelay(){
+				// Todo: Why do we need a popup delay?
```
melink14: It's so if you're just moving your mouse from one place to another, you don't get a ton of japanese pop ups, but if you pause your mouse over a word then the you will see the popup.

options.js
```diff
@@ -1,139 +1,193 @@
...
+				var popupDelay;
+				try {
+					popupDelay = parseInt(document.optform.popupDelay.value);
+					if (!isFinite(popupDelay)) {
+						throw Error('infinite');
+					}
+				} catch (err) {
+					popupDelay = 150;
+					popupDelay: "150";
```
melink14: remove this line.

options.js
```diff
@@ -1,139 +1,193 @@
...
+				chrome.extension.getBackgroundPage().rcxMain.config.kanjiinfo = kanjiinfoarray;
+				chrome.extension.getBackgroundPage().rcxMain.config.lineEnding = lineEnding;
+				chrome.extension.getBackgroundPage().rcxMain.config.copySeparator = copySeparator;
+				chrome.extension.getBackgroundPage().rcxMain.config.maxClipCopyEntries = maxClipCopyEntries;
+				chrome.extension.getBackgroundPage().rcxMain.config.popupLocation = popupLocation;
+
+			}
+
+
+			function getPopUpDelay(){
```
melink14: err, for some reason, this function exists twice?

For some reason, you copied the function into an unrelated for loop. :D

options.js
```diff
@@ -1,139 +1,193 @@
...
+			/*function clicktab(tab) {
+				selectedtab = document.getElementById(tab);
+				// change format of all tabs to deselected
+				// change format of selected tab to selected
+				// hide all tab contents
+				// show selected tab contents
+			}*/
+
+
+			document.querySelector('#submit').addEventListener('click', save_options);
```
melink14: this is also duplicated below, etc.

Probably all of this duplication would be fixed if you closed the for loop above correctly.

options.js
```diff
@@ -1,139 +1,193 @@
...
+
+	for(var i=0; i < document.optform.lineEnding.length; ++i) {
+		if(document.optform.lineEnding[i].value === items.lineEnding) {
+//Todo: Update config
+//TOdo: add the second tobi addition
+// Get the current options
+		
+
+// Get the user input
+			function save_options() {
```
melink14: saveOptions

background.js
deshaun93: > why did you remove this semicolon?

Good catch...

background.js
deshaun93: done.

> Two things:
> 
> 1. We should share this array of option with options.js using a global object like `optionNames`
> 2. We should probably make it easier to read by initializing it with one option name per line in alphabetical order.

Done.

background.js
deshaun93: done.

> You've moved initStorage inside of this callback which is a bit weird. I guess it's only used inside the callback but it does look a bit weird to have the function defined here. I would move it to global scope for consistency.

Done...

background.js
deshaun93: Started this... 
A few more changes I have to make.

options.js
deshaun93: Copy that!

background.js
deshaun93: For now I will revert back to localStorage

deshaun93: Makes sense. I must of carelessly just changed it back when I didn't read all of the code.

background.js
deshaun93: Fixed function headers...

options.js
deshaun93: I removed line 121.

options.js
deshaun93: Removed duplicate... Not sure what happen there

options.js
deshaun93: FIXED THAT WEIRD STUFF!

options.js
deshaun93: Started converting everything to camel case

background.js
```diff
@@ -1,128 +1,374 @@
...
-rcxMain.config.popupDelay = parseInt(localStorage["popupDelay"]);
-rcxMain.config.disablekeys = localStorage["disablekeys"];
-rcxMain.config.showOnKey = localStorage["showOnKey"];
-rcxMain.config.kanjicomponents = localStorage["kanjicomponents"];
-rcxMain.config.kanjiinfo = new Array(rcxDict.prototype.numList.length/2);
-for (i = 0; i*2 < rcxDict.prototype.numList.length; i++) {
-	rcxMain.config.kanjiinfo[i] = localStorage[rcxDict.prototype.numList[i*2]];
-}
-
+/**
```
melink14: What's this version info comment?

background.js
```diff
@@ -1,128 +1,374 @@
...
+				var html = rcxMain.dict.makeHtml(request.entry);
+				response(html);
+				break;
+			case 'switchOnlyReading':
+				console.log('switchOnlyReading');
+				if (rcxMain.config.onlyreading == 'true')
+					rcxMain.config.onlyreading = 'false';
+				else
+					rcxMain.config.onlyreading = 'true';
+				//Todo: Replace this with cloud storage
```
melink14: nice catch.

background.js
```diff
@@ -1,128 +1,374 @@
...
+				console.log('copyToClip');
+				rcxMain.copyToClip(sender.tab, request.entry);
+				break;
+			default:
+				console.log(request);
+		}
+	})
+
+var optionsList = [
+	"copySeparator", // Is this field separator?
```
melink14: Yes, there's a feature where you can save the definitions to clipboard, and this allows you to choose tabs or commas etc.

background.js
```diff
@@ -1,128 +1,374 @@
...
+	"maxClipCopyEntries",
+	"minihelp",
+	"onlyreading",
+	"popupcolor",
+	"popupDelay",
+	"popupLocation",
+	"textboxhl",
+	"showOnKey"];
+
+var kanjiInfoLabelList = rcxDict.prototype.kanjiInfoLabelList;
```
melink14: this initialization section changed a lot more than I thought it would so I'll look more later.

Looks like it may be caused by diffs poor move detection but the file is still double the number of lines...

background.js
```diff
@@ -1,128 +1,374 @@
...
+		localStorage[key] = initialValue;
+		return true;
+	}
+	return false;
+}
+
+/**
+ * Populates config with option values from 3 places
+ * 1. Cloud Storage, 2. Local Storage, and 3. Default
+ * If any values are in local storage they will overwrite cloud storage values.
```
melink14: local storage shouldn't overwrite cloud storage unless cloud storage is empty.

background.js
```diff
@@ -1,128 +1,374 @@
...
+ * Values should no longer be saved to local storage.
+ *
+ * @author Deshaun (From Dev Color)
+ * @param items option values retrieved from cloud storage
+ * @return None
```
melink14: None isn't a return type in Javascript...

instead just omit it!

background.js
```diff
@@ -1,128 +1,374 @@
...
+	return false;
+}
+
+/**
+ * Populates config with option values from 3 places
+ * 1. Cloud Storage, 2. Local Storage, and 3. Default
+ * If any values are in local storage they will overwrite cloud storage values.
+ * Values should no longer be saved to local storage.
+ *
+ * @author Deshaun (From Dev Color)
```
melink14: you don't actually need this since you're submitting directly to the repro and so your authorship is embedded in metadata (unlike above where I just took code from web)

background.js
```diff
@@ -1,128 +1,374 @@
...
+			localStorage['textboxhl'] = "false";
+		if (localStorage['onlyreading'] == "yes")
+			localStorage['onlyreading'] = "true";
+		if (localStorage['onlyreading'] == "no")
+			localStorage['onlyreading'] = "false";
+	}
+
+	//Todo: Need to Organize these options in alphabetical order.
+	/** Set Copy Separator Option value */
+	if (items.copySeparator) {
```
melink14: instead of 3 lines for every I think we can just modify initStorage to do these checks.

We also need to be careful about type differences since local storage is all strings while cloud sync can have real types.

options.html
```diff
@@ -1,148 +1,172 @@
...
-					<div id="gencon" class="tabContent">
-					<h1>General</h1>
-						<p>Popup color:
-							<select class="inputfield" id="popupcolor" name="popupcolor">
-								<option value="blue">Blue</option>
-								<option value="lightblue">Light Blue</option>
-								<option value="black">Black</option>
-								<option value="yellow">Yellow</option>
-							</select>
-							<br />
```
melink14: why did you delete the space after the br tags everywhere?

your editor is out of control.

This file has no functional changes at all does it? it's all just reformat?

General, rule is to not combine refacotring with functional changes since it makes it hard to separate what is important to which.

background.js
```diff
@@ -1,128 +1,374 @@
...
+	}
+
+	if (localStorage["copySeparator"]) {
+		rcxMain.config.copySeparator = localStorage["copySeparator"];
+	}
+
+	if (!rcxMain.config.copySeparator) {
+		rcxMain.config.copySeparator = "tab"
+	}
+	echo(rcxMain.config.copySeparator);
```
melink14: does echo do anything in JS?

background.js
```diff
@@ -1,128 +1,374 @@
...
+	if (localStorage["kanjicomponents"]) {
+		rcxMain.config.kanjicomponents = localStorage["kanjicomponents"];
+	}
+
+	if (!rcxMain.config.kanjicomponents) {
+		rcxMain.config.kanjicomponents = "true"
+	}
+
+	/** Set kanjiinfo option values */
+	rcxMain.config.kanjiinfo = new Array(kanjiInfoLabelList.length / 2);
```
melink14: I haven't looked close yet but I hope we can somehow reuse the logic we have in options.js for this.

background.js
```diff
@@ -1,128 +1,374 @@
...
+
+		if (localStorage[kanjiInfoLabelList[i * 2]]) {
+			rcxMain.config.kanjiinfo[i] = localStorage[kanjiInfoLabelList[i * 2]];
+		}
+
+		if (!rcxMain.config.kanjiinfo) {
+			rcxMain.config.kanjiinfo[i] = "true"
+		}
+
+		// Saving each option value to kanjiInfoObject to be stored in cloud later.
```
melink14: is this required? can't we save rcxMain.config.kanjiInfo?

options.html
```diff
@@ -1,148 +1,172 @@
...
+                Show popup only on pressed key:<br/>
+                <input type="radio" name="showOnKey" value="">Not used<br/>
+                <input type="radio" name="showOnKey" value="Alt">Alt<br/>
+                <input type="radio" name="showOnKey" value="Ctrl">Ctrl<br/>
+                <input type="radio" name="showOnKey" value="Alt+Ctrl">Alt+Ctrl<br/>
+                <br/>
+                <strong>Keys when popup is visible</strong><br><br>
+                <table cellspacing=5>
+                    <tr>
+                        <td>A</td>
```
melink14: why did you add newlines to these options? It seems maybe harder to read than having key -> meaning on the same line?

options.js
```diff
@@ -1,139 +1,141 @@
...
+			document.optform.kanjicomponents.checked = items.kanjicomponents;
+			document.optform.popupDelay.value = items.popupDelay;
+
+			kanjiInfoLabelList = chrome.extension.getBackgroundPage().rcxDict.prototype.kanjiInfoLabelList;
+
+			for (i = 0; i * 2 < kanjiInfoLabelList.length; i++) {
+				// Need to get every other element in the storage, so this funky math was added.
+				// We have abbreviations and values..
+
+				document.getElementById(kanjiInfoLabelList[i * 2]).checked = items.kanjiInfo[kanjiInfoLabelList[i * 2]] //== 'true' ? true : false;
```
melink14: Do we actually need this? can we just loop through items.kanjiInfo?

PR #168, opened 2019-10-18
Add GDocs support and refactor options

mshvern:
### Problem

1. Google docs groups lines in a weird way where the next line is a nextSibling of a DIV. Rikaikun does not acknowledge divs as inline elements, therefore it doesn't scan the next line for the continuation of the phrase.

2. Options code is outdated and needs refactoring. The function responsible for saving the options has a TODO.

### Solution

1. Add a `Compatibility` section to the settings page, under which we include Google Docs support, which is essentially adding `DIV` into the list of inline elements (UGLY, but `false` by default and entirely optional)

2. Refactor the code in options.js and background.js to allow for an easier way to introduce an option.

### Proof

New settings section:
image

Results with the setting **OFF**. 暗くis not recognised and we're suggested a singular kanji.
image
 
Results with the setting **ON**. 暗くis recognised and we're suggested the full word.
image

### DIFF
--- a/background.js
+++ b/background.js
@@ -69,6 +69,7 @@ var optionsList = [
 	"textboxhl",
 	"ttsEnabled",
 	"showOnKey",
+	"inlineDivs"
 ];
 
 /** Get option data from cloud and initialize into memory. */
@@ -118,6 +119,7 @@ function initializeConfigFromCloudOrLocalStorageOrDefaults(cloudStorage) {
 	initConfig('showOnKey', '');
 	initConfig('textboxhl', false);
 	initConfig('ttsEnabled', false);
+	initConfig('inlineDivs', false);
 
 	/**
 	 * Set kanjiInfo option values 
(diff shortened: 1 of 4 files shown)


melink14:
Highlevel thoughts before doing a thorough review:

Is there a more general way to solve this problem that works everywhere? I'm reluctant to add a 'hacks' section to options.

To be clear, the only thing you changed in options was to use rcxMain option directly. That seems better but there are some other open issues to do a larger refactor!

mshvern:
> Is there a more general way to solve this problem that works everywhere? I'm reluctant to add a 'hacks' section to options.

It would be hard to come up with something general here, google docs is just something I use a lot so I needed something for it. 

> To be clear, the only thing you changed in options was to use rcxMain option directly. That seems better but there are some other open issues to do a larger refactor!

Yes, no functionality change, this is just a straight-up refactor, I believe it does improve readability and reduce lines of code, which would make any other open tickets easier to work on.

PR #203, opened 2020-07-27
Update package.json and fix formatting issues in HTML and CSS files.

melink14:
Formatting and fixes was done by expanding the scope of prettier formatter to include HTML and CSS. A few additional CSS lint errors were fixed (reported by vscode).

package.json was updated to be more extensible with separate commands for the base programs that are then combined for the final lint and fix commands.

### DIFF
--- a/extension/options.html
+++ b/extension/options.html
@@ -160,7 +160,6 @@
           </div>
           <div id="clipcon" class="tabContent">
             <h1>Copy to Clipboard</h1>
-            <p></p>
             <table cellspacing="5">
               <tr>
                 <td>Line ending:</td>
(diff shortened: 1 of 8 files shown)


extension/options.html
```diff
@@ -1,155 +1,209 @@
...
+              Tuttle Kanji Dictionary
+              <br />
+              <input type="checkbox" id="U" name="U" />
+              Unicode
+              <br />
+            </p>
+          </div>
+          <div id="clipcon" class="tabContent">
+            <h1>Copy to Clipboard</h1>
+            <p></p>
```
melink14, line 163: Tested in chrome and found that removing this doesn't change spacing at all so will delete.

PR #219, opened 2020-08-21
Format code with ClangFormat.

melink14:
This is a preliminary PR to see if we like the outcome of clang format.

It handles comments by default better than the prettier plugins and is what is used at Google.

The main differences besides when to break lines is 4 spaces for a continuation instead of 2.

If we like this, the follow up will be to update the config files to use this instead of prettier.

---
This change is [Reviewable](https://reviewable.io/reviews/melink14/rikaikun/219)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,169 @@
+---
+Language:        JavaScript
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: AlwaysBreak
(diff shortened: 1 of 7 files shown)


melink14:
Broken so closing.

melink14:
Going to compare it a bit more and hope it's not broken!

melink14:
Decided to go with prettier after https://github.com/hosseinmd/prettier-plugin-jsdoc/issues/5 was fixed after all.

PR #229, opened 2020-08-29
feat(dict): Add Heisig keyword information to kanji dictionary.

melink14:
Added Heisig data parser to update-data.ts and inject the info into the kanji entries.
Data sourced from https://github.com/sdcr/heisig-kanjis/blob/master/heisig-kanjis.csv.
Added an engine to package.json due to `fs.promises` being newer than the default.

Heisig keywords were [originally added](https://github.com/melink14/rikaikun/pull/134) by a contributor as direct edits to the dictionary files and so were lost after switching to the dictionary update script. I noticed the 6th edition indices were missing and easily [added them back](https://github.com/melink14/rikaikun/pull/197) since they were storedin Kanjidcit.

Later, I saw some code referencing Heisig keywords and realized the original PR did more than I thought. This PR restores the complete functionality of #134 via the update script.

This change is [Reviewable](https://reviewable.io/reviews/melink14/rikaikun/229)


utils/update-db.ts
```diff
@@ -350,11 +372,36 @@ class KanjiDictParser extends Writable {
...
+  id_5th_ed: string;
+  id_6th_ed: string;
+  keyword_5th_ed: string;
+  keyword_6th_ed: string;
+  components: string;
+  on_reading: string;
+  kun_reading: string;
+}
+
+const loadHesigData = async (): Promise<Map<string, HeisigData>> => {
```
Stephie, line 386: `loadHeisigData`

utils/update-db.ts
```diff
@@ -350,11 +372,36 @@ class KanjiDictParser extends Writable {
   }
 }
 
+interface HeisigData {
```
Stephie, line 375: Using (H/h)eisigData to refer to all of the data and the format for a single kanji's data is a little confusing.

Stephie (approved):

utils/update-db.ts
melink14: Made singular type HeisigDatum

PR #266, opened 2020-09-06
fix: Remove unneeded tabs permission from manifest.

### LINKED ISSUE #152: Try to remove tabs permission from manifest
It seems like tabs permission is only need to read some special fields of Tab and most chrome.tabs usages don't need it.

Attempt to remove it:
https://developer.chrome.com/extensions/tabs

melink14:
As per https://developer.chrome.com/extensions/tabs the permission
is only needed if accessing sensitive tab information like URL or title.
rikaikun only needs chrome.tabs for sending messages which works fine.

TESTED: Manually tested core behavior.
Fixes #152


Stephie (approved):
🍀

Issue #346, opened 2020-11-30
Options initialization fails for falsy values


melink14:
This only affects boolean values (or numeric values that can take 0) that default to a value other than false but should be corrected to be an explicit check.

> Nevermind, my specific problem is because this line:
> 
> https://github.com/melink14/rikaikun/blob/bb011dc9e825754bef36e89ab5d2cc58416cd2c7/extension/background.ts#L94
> 
> When the stored value is `false`, it'll always fallback to default.
> 
> _Originally posted by @nanaya in https://github.com/melink14/rikaikun/issues/333#issuecomment-735950573_

melink14:
Also from nanaya:

> Yes. Thanks. On newer typescript replacing || with ?? should be enough to fix it.

melink14:
Should also check and test the kanjiinfo settings:
https://github.com/melink14/rikaikun/blob/bb011dc9e825754bef36e89ab5d2cc58416cd2c7/extension/background.ts#L127

PR #408, opened 2021-04-22
fix: Falsy setting values now correctly initialize.

SamDunlap:
I changed the logic that checks whether or not initial options values exist and whether or not they are true, to only check for whether or not they exist already. This way, if a false value for a checkbox is not mistaken for a missing setting value.

Solves issue https://github.com/melink14/rikaikun/issues/346.

--I attempted to rebase these changes as I initially committed without running the linter and I wanted to follow the "one commit per merge req" rule, however this was not recorded by Github.

### DIFF
--- a/extension/background.ts
+++ b/extension/background.ts
@@ -90,8 +90,12 @@ function initializeConfigFromCloudOrLocalStorageOrDefaults(cloudStorage) {
    * @param {boolean | number | string} defaultValue
    */
   function initConfig(key, defaultValue) {
-    let currentValue =
-      cloudStorage[key] || normalizeStringValue(localStorage[key]);
+    const cloudValue = cloudStorage[key];
+    const localValue = normalizeStringValue(localStorage[key]);
+    let currentValue = localValue;
+    if (localValue === undefined) {
+      currentValue = cloudValue;
+    }
     if (currentValue === undefined) {
       currentValue = defaultValue;
     }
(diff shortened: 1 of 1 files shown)


extension/background.ts
```diff
@@ -90,8 +90,12 @@ function initializeConfigFromCloudOrLocalStorageOrDefaults(cloudStorage) {
    * @param {boolean | number | string} defaultValue
    */
   function initConfig(key, defaultValue) {
-    let currentValue =
-      cloudStorage[key] || normalizeStringValue(localStorage[key]);
+    const cloudValue = cloudStorage[key];
```
melink14, line 93: This definitely works but is a bit verbose:

Based on the idea from another user in the issue: https://github.com/melink14/rikaikun/issues/346#issuecomment-736203517

it seems like maybe it might be more idiomatic in typescript to just replace `||` with `??` (assuming it works as described)

Not sure if you already tried it but if not and it works, it seems cleaner!

(reference for nullish coalescing (`??`) operator in TS for FYI: https://medium.com/@sredmond/typescript-tip-of-the-week-nullish-coalescing-vs-logical-or-operator-72779807051)

extension/background.ts
```diff
@@ -124,7 +128,7 @@ function initializeConfigFromCloudOrLocalStorageOrDefaults(cloudStorage) {
   const kanjiInfoLabelList = RcxDict.prototype.kanjiInfoLabelList;
   for (i = 0; i < kanjiInfoLabelList.length; i += 2) {
     const kanjiInfoKey = kanjiInfoLabelList[i];
-    if (cloudStorage.kanjiInfo && cloudStorage.kanjiInfo[kanjiInfoKey]) {
+    if (cloudStorage.kanjiInfo) {
```
melink14, line 131: Oh, this is clever. Since `kanjiInfo` always initialized as a whole we don't need to check the individual values existing!

👍🏾

melink14 (changes requested):
Thanks a lot for this fix.

Don't worry too much about the rebase not working (perhaps the push failed since you didn't pass `--force` as required for rebases) when I merge the PR it will be sqashed to one.

extension/background.ts
SamDunlap: Yeah I hadn't initially tried it, but it does fix the issue that was occurring, so I updated the PR. That operator is so convenient.

extension/background.ts
SamDunlap: Thanks!

melink14 (approved):

melink14:
@all-contributors please add @SamDunlap for code

Issue #230, opened 2020-08-30
Fix typescript compilation errors


melink14:
The files were moved to typescript in #160 but all compiler errors were left in to allow gradual migration.

Errors are suppressed in two places:
eslintrc.json which suppresses many typescript lintings
webpack.config.ts which has a flag for ignoreing errors.

This issue can be incrementally fixed by editing one of those two files and fixing a single class of error, or perhaps fixing a single file of all errors.

One issue is that once you fix a single file it may be hard to stop backsliding unless we can figure out how to turn on strict compilation for just that file.

One first step might be to define classes and modules for the background scripts such that they don't rely on global state since that will affect all files.

melink14:
I'm going to start with just connecting them to each other so we're not relying on globals.

melink14:
webpack stopped outputting es6 modules so I may have to switch to rollup.

melink14:
Though it does appear to still work so maybe I'll just roll with it even though it's not as efficient.

PR #430, opened 2021-05-02
refactor(data): Fixes compiler warnings in data.js

melink14:
- Creates proper types and interfaces for various dictionary data.
- Simplies dictionary loading so that types are consistent and readable.

Contributes to #230.

### DIFF
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -449,11 +449,16 @@ class RcxDict {
             cache[u.word] = [];
             continue;
           }
-          ix = result.split(',').map(parseInt);
+          // The first value in result is the word itself so skip it
+          // and parse the remaining values at integers.
+          ix = result
+            .split(',')
+            .slice(1)
+            .map((offset) => parseInt(offset));
           cache[u.word] = ix;
         }
 
-        for (let j = 1; j < ix.length; ++j) {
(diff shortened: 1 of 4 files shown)


extension/data.ts
```diff
@@ -43,16 +43,58 @@
...
   nameDict?: string;
   nameIndex?: string;
-  difReasons: never[];
-  difRules: never[];
-  difExact: never[];
-  radData: string;
+  difReasons: string[] = [];
+  difRules: DeinflectionRuleGroup[] = [];
+  radData: string[] = [];
+  wordDict = '';
```
melink14, line 95: group dictionary variables together?

extension/data.ts
```diff
@@ -63,40 +105,49 @@ class RcxDict {
     return RcxDict.instance;
   }
 
-  init(loadNames) {
-    const started = +new Date();
+  static createDefaultDictEntry(): DictEntryData {
```
melink14, line 108: Try to remove this by making one at module scope and deriving type from it.

extension/data.ts
melink14: done.

extension/data.ts
melink14: Some values were optionally null but I just changed their defaults to falsy string/number which should work.

PR #431, opened 2021-05-02
ci: Add explicit step to report comment status to PR.

melink14:
Due to new security measures updating the status explicitly won't work from forks so we need to use workflow run as per: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

This can't be tested before merging so it may change later.

### DIFF
--- a/.github/workflows/update-pr-status.yml
+++ b/.github/workflows/update-pr-status.yml
@@ -9,11 +9,9 @@ jobs:
   updatePRStatus:
     runs-on: ubuntu-latest
     steps:
-      - run:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-pr-status.yml
```diff
@@ -0,0 +1,26 @@
+name: Update PR status
+on:
+  workflow_run:
+    workflows: ['No unresolved comments']
+    types:
+      - completed
+
+jobs:
+  updatePRStatus:
+    runs-on: ubuntu-latest
```
melink14, line 10: Does this actually work?

PR #454, opened 2021-05-06
chore(build): Move heisig-kanjis.csv out of extension directory since it's only used to build kanji.dat

### LINKED ISSUE #267: Move heisig-kanjis.csv out of extension directory since it's only used to build kanji.dat
No need to add 254 KiB to the extension since it's already heavy.

melink14:
Fixes #267

TESTED=Ran `npm run update-db` and it finished fine!


Stephie (approved):
Okay!

PR #458, opened 2021-05-06
docs(readme): Make all badges the same style (flat) and add relevant links for users/license badges.

melink14:
- Chrome web store users badge now links to chrome web store
- license badge links to license file


Stephie (approved):
Yay!

PR #462, opened 2021-05-06
fix: Reset badge text to empty on extension startup

### LINKED ISSUE #82: "On" icon incorrectly displayed on startup
```
What steps will reproduce the problem?
1. Open chrome for first time
2. "On" icon displayed
3. Confirm that rikaikun is not actually on
4. Press the rikai button again to turn on
5. Observe that icon remains in "on" state and rikai kun is actually on

What is the expected output? What do you see instead?
Expect the icon to not display on when rikai kun is not on


What version of the product are you using? On what operating system?
 0.8.5
on chrome 22.0.1229.94 m
and windows 7 x86
```

Original issue reported on code.google.com by `perceptu...@gmail.com` on 15 Oct 2012 at 4:22

melink14:
For some reason, chrome restores the previous badge state on restart causing inconsistency with actual rikaikun state.

TESTED=Loaded unpacked extension and used chrome://restart to ensure badge text reset.
Fixes #82

### DIFF
--- a/extension/background.ts
+++ b/extension/background.ts
@@ -47,6 +47,10 @@ chrome.runtime.onMessage.addListener(function (request, sender, response) {
   }
 });
 
+// Clear browser action badge text on first load
+// Chrome preserves last state which is usually 'On'
+chrome.browserAction.setBadgeText({ text: '' });
+
 rcxMain.config = {};
 
 const optionsList = [


Stephie (approved):
Okay!

Later, it might be worth making a boolean badge state instead of setting the badge text directly, but this is fine for now.

PR #488, opened 2021-05-20
refactor(options): Use advanced type guards to confirm types when saving generic options

melink14:


melink14:
This and related PRs were abandoned in favor of smaller handlers called directly on the HTML.

PR #516, opened 2021-05-27
refactor(rikaicontent): Fix compilation errors in rikaicontent.ts

### LINKED ISSUE #230: Fix typescript compilation errors

melink14:
- Fix type issues without changing logic in rikaicontent.ts
  -  Later, need to fix loose types and linter ignores.
- Fix minor issue in texttospeech.ts since it was the only one left.
- Disable unbound-method check in options LitElement since it's not lit aware and should be generally safe.

Fixes #230

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,6 +19,7 @@
     "@typescript-eslint/no-var-requires": "off",
     "unicode-bom": "error",
     "promise/prefer-await-to-then": "error",
+    "@typescript-eslint/unbound-method": "error",
     // TODO(espeed): Remove these and fix errors.
     "no-undef": "off",
     "prefer-arrow-callback": "off",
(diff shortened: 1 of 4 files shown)


melink14:
Compiler fixes so skip approval.

PR #519, opened 2021-05-28
chore(build): Change typescript loader to `ts-loader`

melink14:
awesome-typescript-loader was good for allowing builds ignoring errors but is unmaintained. Now that compilation passes we can switch to a modern alternative.
Also, removes performance hints from webpack build since they're noise for a chrome extension.


Stephie (approved):
Okay.

PR #521, opened 2021-05-28
refactor(background): Change `initialConfig` -> `config` in all remaining places

melink14:
- `rename variable` is scope specific it turns out.


Stephie (approved):
Ok!

PR #523, opened 2021-05-28
refactor(fix-lint): Turn on and fix simple lint checks

melink14:
Enables any checks that could be fixed with small, mechanical improvements to the code base.


extension/rikaicontent.ts
```diff
@@ -902,6 +901,7 @@ class RcxContent {
       return;
     }
 
+    // TODO(melink14): Why does this replace all characters with HTML escape codes
```
Stephie, line 904: 😂

Stephie (approved):

PR #525, opened 2021-05-28
refactor(fix-lint): Enable eqeqeq check and fix errors

melink14:
- Most changes can mechanically be changed to the proper operator
- Some null checks were ambiguously comparing against undefined and
  needed to be fixed.


Stephie (approved):
I trust you!

PR #561, opened 2021-06-26
feat(detection): Ignore invisible nodes when extracting text under mouse

### LINKED ISSUE #159: Decide and implement unit testing framework
Probably use Jasmine with karma?

At first should add test harness and a couple of tests as proof of concept.

(maybe add some incremental coverage measurements)

### LINKED ISSUE #366: Invisible HTML prevents word detection
rikaikun tries to flatten dom but in this case some weird auxillary hidden text has been added to facilitate some minor mini dictionary. Since they are considered text nodes, even if invisible they interrupt the word.

Part of this is the netflix language extension's poor grammar parsing but we could also try to ignore invisible elements...

```html
<span data-token-index="6" data-word-key="WORD|絞る|ja" class="lln-word lln-hover-tooltip top">
  <span class="tt lln-tt-hidden" style="font-size: 21.8636px;">
    <span style="white-space: nowrap; opacity: 1;">squeeze</span><br />
    <span style="white-space: nowrap; opacity: 0.8;">wring</span>
  </span>
  絞っ
</span>
<span data-token-index="7" data-word-key="WORD|て|ja" class="lln-word lln-hover-tooltip top">
  <span class="tt lln-tt-hidden" style="font-size: 21.8636px;">
    <span style="white-space: nowrap; opacity: 1;">have</span>
  </span>
  て
</span>
```

melink14:
Specifically, ignore `display: none` and `visibility: hidden`.

This change is also the first unit tested change so contains the initial
configuration of testing. (See #159 for details on decisions).

Testing Stack: @web/test-runner + Mocha/Chai/Sinon + Puppeteer Headless Chrome.
Works with vscode debugging for fast test driven development.

Fixes #366
Fixes #159


melink14:
Will completely change this to be based on snowpack soon!

extension/rikaicontent.ts
```diff
@@ -679,6 +685,11 @@ class RcxContent {
     return text;
   }
 
+  #isElementVisible(element: HTMLElement) {
```
melink14, line 688: Using JS private field might not make sense in typescript since `private` will prevent developers from doing things accidentally and prod code doesn't really need to be so strict.

extension/rikaicontent.ts
```diff
@@ -1286,3 +1297,5 @@ chrome.runtime.onMessage.addListener((request) => {
 
 // When a page first loads, checks to see if it should enable script
 chrome.runtime.sendMessage({ type: 'enable?' });
+
+export { RcxContent as TestOnlyRcxContent };
```
melink14, line 1301: I set it up so this gets stripped in prod (since it actually breaks prod); but I wonder if I should add a comment. (Maybe in the future ic could use something generic like `// TESTONLY: <desc>` which will cause the next line to be stripped.

extension/rikaicontent.ts
melink14: Changed it typescript private only!

extension/rikaicontent.ts
melink14: Comments are hard to use actually because esbuild strips all of them. So far going forward it will proably work just to use TestOnly prefix in a identifier on the line somewhere.

PR #598, opened 2021-07-15
fix: Skip text processing when range is null

### LINKED ISSUE #386: Remove console logs from production build
Would it be possible to remove all `console.log()` calls in the production build? The extension frequently logs errors to the console and as a dev, it's distracting. Most commonly, this part of the code throws an error when "x or y are negative, outside viewport, or there is no text entry node" and range ends up null. For me, mousing over the scrollbar often triggers this. I'm using Chrome 89 though this issue has been present on previous versions as well.

```
range = document.caretRangeFromPoint(ev.clientX, ev.clientY);
rp = range.startContainer;
```

image

Really appreciate the work on this extension :)

qkjosh:
Avoid raising errors when range is null. Calling caretRangeFromPoint when "x or y are negative, outside viewport, or there is no text entry node" will return null. One common example is hovering over a scrollbar.

Fixes #386


melink14:
This looks good and thanks for the contribution! I'll probably wait til I merge the #561 PR statck since that adds testing framework that I want to start using for incoming changes now that I finally have all TS and lint errors fixed. (I got behind on code reviews so the stack has been growing...)

I'll probably do that tonight or this weekend JST and then add a quick test to your PR and merge it.

If you're interested you could retarget your PR to that branch chain and then add a quick test yourself after looking at the examples in my PRs but otherwise merging shouldn't take too long I hope (famous last words)

melink14:
I caught up on pull requests for now so I'll check this out in more detail this weekend and then merge dictionary PR which will trigger store update!

I might want to keep the larger console issue open since it occurs to me that instead of printing client errors directly it would be better to report them to the background script for printing in that debug console. (assuming it still works with MV3)

qkjosh:
Thanks! Not in any big rush, whenever you can get to it is fine. Unfortunately I don't think I'll have time to add unit tests in the next couple days, before you get around to reviewing/merging it anyways. If this is still open after the weekend I can take a stab at it though, for sure. I think it should be enough to test the `tryUpdatePopup` function, passing it a MouseEvent with negative clientX/clientY values.

melink14:
I added a test for the branch early return but coverage is penalizing me for not testing that the inverse of the branch as well. I'd like near 100% relative coverage if possible but in this case the missing branch is the happy path and so adding a test would be more code health work so I'll probably just leave it for now.

melink14 (approved):

melink14:
@all-contributors please add @qkjosh for bug report and fix.

melink14:
(looks like natural language parsing wasn't good enough)
@all-contributors also add @qkjosh for code.

PR #933, opened 2022-04-02
feat(ui): Isolate rikaikun popup from host page via shadow DOM and `all: revert`

### LINKED ISSUE #221: Encapsulate rikaikun popup in shadowdom to prevent weird interactions with host page
Currently, there's a CSS reset in rikaikun to avoid most issues but using shadow dom should be more scalable.

It also will allow us to simply CSS since it will be scoped.

References:
https://developers.google.com/web/fundamentals/web-components/shadowdom
Maybe useful example: https://blog.railwaymen.org/chrome-extensions-shadow-dom

melink14:
- Creates a div inside of shadow root to contain all styles and modifications.
- Updates CSS to use `all: revert` instead of previous adhoc reset.
- Installs CSS directly into the shadow DOM as per best practices.
- Fallsback to `all: initial` for older Chrome browsers.
- Removes special check for text files since chrome renders them as normal HTML pages.

Fixes #221


melink14:
Will ignore codecov this time since it's caused by variable name change only and the previous lines were also uncovered.
percy doesn't work at all with shadow dom so I guess I'll disable it for now: percy/cli/issues/280

PR #960, opened 2022-04-07
fix(ui): Reset styles on the light dom container

### LINKED ISSUE #959: Pop-up hidden if host page styles div as `position: relative`
This was reported via CWS as seen on instagram.

I as thinking that the outer div didn't matter but of course the absolute positioning of the shadow popup is affected if it's inside a relative positioned parent.

I think setting resetting all styles on the shadow host to:
`all: revert !important` should be good to ensure we start with a blank slate for non inheritable properties.

melink14:
- Add an important `all: revert` to the light dom style attribute
- Hide and show popup based on display of shadow container instead of light container. This prevents conflicts with light container reset and is more resiliant to changes to the light container.
- Adds new condition to ignore styles test and a test which ensures visibility check is accurate.

Fixes #959


melink14:
Absolute code coverage changes are odd but since I only added test coverage I think any changes must be coincidental!

PR #962, opened 2022-04-07
fix(ui): Make Google Doc Canvas checks more lenient

### LINKED ISSUE #881: Rikaikun doesn't work in Google Docs for some users (TrustedScript violation?)
From the support tab:
> When I bring up developer tools the error messages are "VM290:27 This document requires 'TrustedScript' assignment." and "Trusted Type expected, but String received"

Maybe we can created a `TrustedScript` but policy creation can also be hardened...

Can check if google docs is running trusted types experiment...

refs:
https://web.dev/trusted-types/

### LINKED ISSUE #897: Not detecting text in scrollable section of google doc
There's some weird HTML covering the japanese text which is not accounted for in the current implementation:

```html
<div class="kix-domviewscroller-outer docs-ui-unprintable __web-inspector-hide-shortcut__" style="left: 40px; top: 21778px; height: 59px; width: 532px;"><div class="kix-domviewscroller-inner" style="cursor: text; width: 532px; height: 59px;"></div></div>
```

Might consider just not being so strict about what type of dom element we're colliding with since it's probably not slowing anything down anyway if we're already in a doc.

melink14:
I found in offline mode I the cached Google Docs query selector was always false due to executing before page loaded. I now query the selector on every `mousemove` which seems fast enough. An improvement would be to memoize a function or perhaps add a on Dom loaded callback.
I also removed the check for if the target was `svg` or `rect` since there's at least one counter example to that assumption. For now let's avoid premature optimization.

I also had to update the tests to not overwrite the Google Docs class since it now needs to be present at each execution.

Fixes #897
Fixes #881

### DIFF
--- a/extension/rikaicontent.ts
+++ b/extension/rikaicontent.ts
@@ -73,9 +73,6 @@ class RcxContent {
   private defaultDict = 2;
   private nextDict = 3;
 
-  private isGoogleDocPage =
-    document.querySelector('.kix-canvas-tile-content') !== null;
-
   // Adds the listeners and stuff.
   enableTab(config: Config) {
     if (window.rikaichan === undefined) {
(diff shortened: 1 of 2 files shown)


melink14:
Decrease in coverage is due to total number of lines decreasing.

PR #988, opened 2022-04-22
feat(deinflect): Add logic to handle all inflections of the copula だ

### LINKED ISSUE #89: でした not recognized
```
What steps will reproduce the problem?
1.It doesn't recognize "でした" (was/were);
2.In "Options", only "General" is active. 


What version of the product are you using? On what operating system?

Version 0.8.5, on Google Chrome.



```

Original issue reported on code.google.com by `ciccio.d...@gmail.com` on 23 Jul 2013 at 12:44

maawisul:
- Adds a new type of deinflection for the copula, だ.
- Adds mappings for all common conjugations to deinflect.dat.
    - See http://www.japaneseprofessor.com/reference/grammar/conjugations-of-the-japanese-copula/
- For handling polite negative じゃないです update general adjective handling to recognize polite です addition.

Fixes #89


melink14:
I think the change to make it conjugate as a verb might be too broad since it will cause it to match other verb endings as well:
image

I wasn't thinking it through when I thought we could reuse the existing す verb logic... 🙇🏾 

Since です is so irregular I think the complete fix would be to add a new type for `cop` in deinflect.dat and then add entries for the cases which only apply to this type of word. (https://www.tanoshiijapanese.com/dictionary/conjugation_details.cfm?entry_id=103243&element_id=128958&conjugation_type_id=5 is the best list I found but we don't need to add them all now.)

To add an entry to deinflect.dat is faily straight forward. It starts with a set of reasons and then each entry has the form of:
`<inflected_form> <deinflected_form> <type_mask> <index_into_reasons>`

The first two are encoding the logic we want and the last is a 0 based index into the reasons at the top of the file.

The type mask is opaque but it looks to be two values OR'd together (type of inflected word and type of deinflected word) as was clarified in this fork: https://gist.github.com/lynn/07f7ce3c314223d2aca19ec2bb0540cd#file-patch-deinflect-py-L7

I think if we add a new base mask for です then we can easily add a line to deinflect.dat and update your branch a bit to make things work specifically for this case!

What do you think?

maawisul:
I agree! I did think that the case of です and its conjugation is pretty irregular and need to be differentiated from any other verbs or adjectives. However, I am not sure on how to come up with a new typemask and would need some guidance on that.

melink14:
No problem! I have some ideas about that.

For the type of the inflected word, I think 0x0080 (OTHER) makes sense in this case (though if we added forms like ではない it would be different). For BASE we'll need a new type in the upper half of the mask (imagine `BASE_DESU`) and since `BASE_SURU` is 0x1000 we can increment it to 0x2000 and be safe. (This will convert to a type of 32 in data.ts).

```bash
> node -p "0x80|0x2000"
8320
```

So I think a type mask of 8320 should work well for でした　ー＞　です

Let me know if you have any other questions.

melink14:
BTW, I was reading http://www.japaneseprofessor.com/reference/grammar/conjugations-of-the-japanese-copula/ which made me realize the honorific forms are also lost in rikaikun.

I think we'll be able to add several new lines for various copula conjugations once the base logic is in. You could do it at the same time or just focus on でした depending on your time/want.

maawisul:
Thanks! I will try working on it as much as I can submit another PR when I'm satisfied with my change.

tentagul (approved):

tentagul (approved):

tentagul (approved):

tentagul (approved):

melink14:
> Thanks! I will try working on it as much as I can submit another PR when I'm satisfied with my change.

For now, I'll close this PR then and feel free to re-open when ready, or let me know if you have any more questions on the issue.

maawisul:
I think I am satisfied with my changes and ready to re-open.

maawisul:
I added entries and type(past negative) to deinflect.dat so that rikaikun now recognizes:
でした
じゃありません
じゃありませんでした
でございません / ではございません
でございませんでした / ではございません

I am not sure on what to do with ございます on its own as I am not sure if it is still considered a copula or not

melink14:
Thanks for the update! Looks like we're pretty close.

I think maybe things became complicated because the dictionary contains many congugated forms of the copula which you tried to work with directly and thus created congugation mappings to intermediate inflections.

It makes sense given that the dictionary contains things like じゃない even though that's the negative form of だ but I think we should maintain consistency and conjugate back to the base form in all cases. This means that we'll show two entries for some words but that's okay in my mind because sometimes these short common phrases have idiomatic meaning as well as the basic grammar meaning.

I think not relying on the dictionary containing certain common inflections is also more robust since they might remove the inflected forms in a future update.

In that case, it makes sense to add mappings back to だ since it's the polite affirmative form of the copula. We can start with the mappings that don't have any other entry in the dictionary or fill out the entire table. I'll leave a comment with what I mean inline.

extension/data/deinflect.dat
```diff
@@ -333,4 +336,11 @@ imperative negative
 り	る	640	24
 れ	る	640	23
 れ	れる	384	24
-ろ	る	384	23
\ No newline at end of file
+ろ	る	384	23
+した	す	8320	14
```
melink14, line 340:
```suggestion
でした	だ	8320	6
```

melink14: I think if we map like this (similar to how most verbs and adjectives are mapped) then we shouldn't need to make additional changes outside the new mappings and we can add as many or few as we want to fill the holes you want!

maawisul: I see. So instead of conjugating でした to です, I should conjugate it to だ? And also do this for every instances of だ?
What about inflection of でございます?　Should I considerでございます as the base form as I believe でござる is not commonly used.

maawisul: Or should I deinflect でございます back to だ and add a new 'honorific' category in deinflect.dat?

melink14: Yes, you understood it well. This is definitely a special case and there's room for iteration.

One interesting idea from @Stephie was to add mappings to the both だ and です so that people could see both connections. (and similarly maybe have honorific forms go to both だ and でございます)

I think honorific forms usually are treated separately so it's I think it's a fair point to leave it as is. I also agree that deinflecting it to でございます makes sense as the plain form is archaic.

To summarize the simplest approach I think we should take for now:
- です forms go to だ
- でございます forms go to でございます

One thing to note is since these are irregular inflections you should map the entire word instead of just the suffix like you've done till now. (simmilar to how する is treated I think) 
There's no reason to be generic since each line only captures a single mapping as far as I can tell.

maawisul: So, I tried deinflecting to だ but when I tested it, the words are once again not recognized. I feel like it has something to do with this line of code in the deinflection implementation.
https://github.com/melink14/rikaikun/blob/69c7fe799955df76c94e9686ea08ad0b8503e9ae/extension/data.ts#L281
Am I right? If this is really the case, then we can not use だ as the base form to deinflect to since the length is always going to be 1, is that the case?

maawisul: I went with option 1 and did not find any instances where rikaikun produces any weird output yet. I think I would prefer going with option 1 for now.

extension/data/deinflect.dat
```diff
@@ -333,4 +336,34 @@ imperative negative
 り	る	640	24
 れ	る	640	23
 れ	れる	384	24
-ろ	る	384	23
\ No newline at end of file
+ろ	る	384	23
+です	だ	8320	13
+だった	だ	8320	14
+でした	だ	8320	6
+ではない	だ	8320	15
```
melink14, line 343: This hadn't come up before but since this ends in ない that means it doesn't count as `OTHER` anymore and instead counts as an i-adjective. (which has type mask 0x0004). According to my quick calculation that means this type mask should be 8196. I think if you make this change for every mapping that ends in ない then you don't need to add the inflections of ない directly.

```suggestion
ではない	だ	8196	15
```

extension/data/deinflect.dat
```diff
@@ -333,4 +336,34 @@ imperative negative
...
+じゃない	だ	8320	15
+ではないです	だ	8320	1
```
melink14, line 345: I noticed that rikaikun doens't usually include です ending for polite negative (熱くないです for example) since I guess people should know that the です is connected.

That being said, this seems useful for new learners so I think it's fine to keep (and maybe think of extending this logic to other adjective types later)

extension/data/deinflect.dat
```diff
@@ -333,4 +336,34 @@ imperative negative
...
+じゃないです	だ	8320	1
+ではありません	だ	8320	1
+じゃありません	だ	8320	1
+ではなかった	だ	8320	28
```
melink14, line 349: Continuing from above, I think you can remove these two lines if you change the type mask for the root above.

melink14 (changes requested):
This looks pretty good with some small comments.

We should also add some tests for the data.ts changes. There aren't tests for that file yet but it should be fairly straightforward to do in the same manner of other files like background.ts.

That being said, it can be tricky writing the first tests for a class, so let me know if you want me to add a commit with some tests this time around.

extension/data/deinflect.dat
```diff
@@ -123,6 +124,8 @@ imperative negative
 かせる	く	513	9
 がせる	ぐ	513	9
 かった	い	1152	14
+ありません	ない	1152	13
```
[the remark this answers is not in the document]
maawisul: I just thought that it would be nice if the user can see that じゃありません and ではありません comes from じゃない and ではない first and can also see that they are also inflections of だ.

extension/data/deinflect.dat
```diff
@@ -333,4 +336,34 @@ imperative negative
...
+で	だ	8320	21
+でありまして	だ	8320	21
+なら	だ	8320	17
+ならば	だ	8320	17
+であれば	だ	8320	17
+でございました	でございます	8320	14
+ではございません	でございます	8320	15
+ではございませんでした	でございます	8320	28
+でございませんで	ございます	8320	15
+でございませんでした	でございます	8320	28
```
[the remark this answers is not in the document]
maawisul: Yes, I am just not sure if that is the only negative form of the honorifics, I will get rid of the 2 lines.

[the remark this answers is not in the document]
melink14: Yeah, I couldn't find a reference but if it comes up later we can easily add it at that time.

Thanks.

extension/data/deinflect.dat
maawisul: I added an entry so that です is considered a polite inflection for i-adjectives. This does not work when the adjectives are in past form, however, which I believe is due to a different type mask. So, I added another separate entry instead.

extension/data/deinflect.dat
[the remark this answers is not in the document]
maawisul: I removed the 2 lines

extension/data/deinflect.dat
```diff
@@ -333,4 +336,34 @@ imperative negative
...
+でしょう	だ	8320	2
+で	だ	8320	21
+でありまして	だ	8320	21
+なら	だ	8320	17
+ならば	だ	8320	17
+であれば	だ	8320	17
+でございました	でございます	8320	14
+ではございません	でございます	8320	15
+ではございませんでした	でございます	8320	28
+でございませんで	ございます	8320	15
```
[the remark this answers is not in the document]
maawisul: I made a typo and did not realize it. I am sorry

extension/data/deinflect.dat
```diff
@@ -142,6 +142,8 @@ past negative
 きます	くる	2176	13
 ぎます	ぐ	640	13
 くない	い	1028	15
+です		1028	13
```
melink14, line 145: 1028 is probably not the right one for these two since it means i-adj -> i-adj and we want one which is OTHER -> i-adj and that looks like 1152 based on other examples. 

```suggestion
です		1152	13
```

extension/data/deinflect.dat
```diff
@@ -142,6 +142,8 @@ past negative
...
+かったです	い	1028	6
```
melink14, line 146:
```suggestion
かったです	い	1152	6
```

extension/data/deinflect.dat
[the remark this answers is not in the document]
melink14: Actually, in that case can we make these more specific by including the じゃ and では in the mappings.

If I hover over ありません now it has some strange inflections for cases which don't make sense I think (adjectives which are just ない for example)

melink14 (changes requested):

extension/data.ts
```diff
@@ -278,7 +278,7 @@ class RcxDict {
             if (type & rule.typeMask && end === rule.from) {
               const newWord =
                 word.substr(0, word.length - rule.from.length) + rule.to;
-              if (newWord.length <= 1) {
+              if (newWord.length <= 0) {
```
melink14, line 281: Copying from my non review comment so it's easier to notice:

> We should also add some tests for the data.ts changes. There aren't tests for that file yet but it should be fairly straightforward to do in the same manner of other files like background.ts.
> 
> That being said, it can be tricky writing the first tests for a class, so let me know if you want me to add a commit with some tests this time around.

melink14 (changes requested):

extension/data.ts
maawisul: Yes, I would like some help for a commit with some tests as I am not entirely sure on how to do that, thanks!

extension/data/deinflect.dat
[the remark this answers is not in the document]
melink14: (I updated this comment yesterday but forgot to unresolve it. sorry!)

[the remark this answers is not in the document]
maawisul: Ah, I see. I will fix it immediately

extension/data.ts
melink14: I added some tests for the new branches as per the latest commit. The first tests for a method are always a bit tricky so I had to install some new assertion helpers to write tests to the level of readability and robustness that I wanted.

Please read them for your own reference; will definitely be useful if you make other changes in the future! (And also useful as a general software development example.)

melink14:
I think it should be basically done now though it looks like I need to increase the timeout for the new data tests (loading the data is slower in github than on my local computer...)

melink14 (approved):

melink14:
@all-contributors Add @maawisul for code

PR #1089, opened 2022-07-02
fix: Use `fetch` instead of XHR for reading dictionary files

melink14:


melink14:
Absolute coverage went down due to number lower total number of lines. Safe to ignore.

PR #1428, opened 2023-02-26
Fix hiragana lookup arrays

tora-pan:
Upon creating a separate branch to expand code coverage, I found a failing test when working with the hiragana conversion tables. 

Found that the array of characters containing the characters such as `ざじずぜぞ`... had katakana characters `ナニヌネノ`as well as `ヴエイ`.

Converted the　`ヴ`to `ゔ` and removed the `ナニヌネノ`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ If you get stuck, push your code early and ask for feedback on the issue.
 
 1. After you check out the code, run `npm install` ([npm installation guide](https://www.npmjs.com/get-npm)) to get the rikaikun dev tools ready.
    1. By default, `npm install` will also run a test suite and install visual test baselines in the local-screenshots directory.
-2. Make your changes and commit them locally. Run `npm run fix` often to ensure your code follows style guidelines.
+2. Make your changes and commit them locally. Run `npm run fix` often to ensure your code follows style guidelines. This runs before you push by default as well.
 3. Add tests for your new code in the `extension/test` directory.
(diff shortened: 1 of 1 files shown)


melink14:
Thanks @tora-pan for the PR! It's much appreciated.

Can you tell me more about the tests that failed when you found these problems? Were there certain words which weren't found that should have been?

Most of the word search code pre-dates me so I had to study it a bit and it seems that maybe some of the mismatching inclusions were added on purpose and not documented.

For example, I see a comment at the top of word search that katakana ヴ (vu) is never converted to hiragana for some reason. In that case, it makes sense that it would exist in the conversion table.

I think we shouldn't just make this change because of the apparent intent of these tables but instead only fix it if we find actual full words which are broken  due to this problem. I'm worried that some of the apparent odd inclusions actually make various look ups work better.

I'd guess some of this could be answered by looking at the dictionary updating/loading code to see what type of normalization they do...

That said, there's a lot of room for improvement here. The variable names are terrible and the logic should probably be broken up into many helper functions...

melink14:
Oh, I just realized one reason for extra characters would be that the table look up is done using simple index lookup math. So it takes the incoming half width character and subtracts the base unicode character value to get an index to lookup into the table. This means it's very important where in the table certain characters appear. For example, in your new change, we're mapping ｴ to が since that is what appears in index 1 of the table now.

Needless to say this is quite a hard to read and maintain implementation. I think a method which uses a switch or a Map would be better, and then we could use the actual characters instead of unicode numbers given modern IDEs support many character sets.

tora-pan:
Oh, awesome.

Once I wrap up the `data.ts` test coverage branch I'm working on, I'll look into reimplementing that section. I totally agree with you and would like to make it readable for the average Joe.

PR #1436, opened 2023-02-28
refactor(data): Refactor `wordSearch` and other methods for clarity and maintainability

tora-pan:
## Refactor `wordSearch` to use lookup object
- Added a lookup object to more cleanly convert from カタカナ to ひらがな.
- Moved various character codes into constants for better readability.
- Removed nested if/for block and replaced with a more concise/understandable `normalizeInputForLookup` method.

- Renamed a few variables for code readability. 
- (melink14) Made method visibility explicit to avoid testing methods via internal implementation details.

## Testing 
- Cherry picked tests from separate branch to cover newly added methods and to make sure there were no regressions.
- (melink14) Added translate tests to cover true word length tracking logic.


tora-pan:
@melink14 Don't mind the extra commits at the end. I accidentally did a `git add .` and pushed. (pushing my `web-test-runner-config.js` and some .vscode settings.

I did wan't to ask you about that. I have to update my `executablePath` to make sure pupetteer works correctly. Do you think I should just add it to my global `.gitignore` or something?

Let me know what you think about the contributions and please let me know if there is anything that you think could be done differently, refactored, etc...

ありがとうございます！

extension/data.ts
```diff
@@ -63,7 +63,7 @@ type DictEntryData = typeof defaultDictEntryData;
 interface Deinflection {
   word: string;
   type: number;
-  reason: string;
+  reason: string | null;
```
melink14, line 66: I've not looked closely yet but I'm always skeptical about adding nullable types. For example, strings can usually be empty with similar meaning as null.

extension/test/data_test.ts
```diff
@@ -63,4 +64,145 @@ describe('data.ts', function () {
...
+        eigo: 'day, sun, Japan, counter for days',
+      });
+    });
+
+    it('should set kanji property of DictEntryData object', function () {
+      const result = rcxDict.kanjiSearch('日');
+      expect(result?.kanji).to.equal('日');
+    });
+
+    it('should set misc property of DictEntryData object', function () {
```
melink14, line 91: It's a bit more code but I think it'd be more maintainable to split this test up for each type of field. Since in the future if we remove or add fields we'll be able to document it well via test addition or removal.

extension/test/data_test.ts
```diff
@@ -63,4 +64,145 @@ describe('data.ts', function () {
...
+      const result = rcxDict.kanjiSearch('日');
+      expect(result?.misc).to.deep.include({
```
melink14, line 93: Style preference:
Try to separate tests into 2 or 3 blocks:

```
Arrange

Act

Assert
```

With a newline between each one (somtimes arrange won't exist or act and assert will be grouped).

Used consistently it makes it very easy to understand tests we read for the first time.

melink14 (changes requested):
Based on the new screenshots that were pushed it looks like there are some unexpected changes to how words are looked up (less words are found); we'll need to debug that.

extension/data.ts
tora-pan: Looks like this was the main cause of the issue. Thank you for pointing it out.
Removed the `| null` and flipped the logic to check if the character isKana or not.

extension/test/data_test.ts
tora-pan: 了解です。Copy that.

extension/data.ts
tora-pan: I've actually added this back as there is a check towards the end that returns `null` that was present in the original legacy code. Let me know if there is another way you'd like to handle this.

extension/data.ts
```diff
@@ -296,93 +287,308 @@ class RcxDict {
...
-      }
-
-      reason += String.fromCharCode(u);
-      // need to keep real length because of the half-width semi/voiced
-      // conversion
-      trueLen[reason.length] = i + 1;
-      p = v;
+    let isKana = false;
+    for (let i = 0; i < word.length; i++) {
+      isKana = this.isKana(word[i].charCodeAt(0));
```
melink14, line 586: This is a bit weird. Essentially, isKana will always be whatever the last character is.

Should it be kana if all characters or kana or if any character is kana?

In that case, something like `Array.from(word).all(this.isKana))` would probably be clearer. (Didn't test that code so might not be exact)

extension/data.ts
```diff
@@ -417,7 +623,7 @@ class RcxDict {
 
     while (word.length > 0) {
       const showInf = count !== 0;
-      let trys;
+      let trys: Deinflection[];
```
melink14, line 626: maybe trys -> possibleDeinflections would be good.

extension/data.ts
```diff
@@ -296,93 +287,308 @@ class RcxDict {
...
-  ch: number[] = [
-    0x3092, 0x3041, 0x3043, 0x3045, 0x3047, 0x3049, 0x3083, 0x3085, 0x3087,
-    0x3063, 0x30fc, 0x3042, 0x3044, 0x3046, 0x3048, 0x304a, 0x304b, 0x304d,
-    0x304f, 0x3051, 0x3053, 0x3055, 0x3057, 0x3059, 0x305b, 0x305d, 0x305f,
-    0x3061, 0x3064, 0x3066, 0x3068, 0x306a, 0x306b, 0x306c, 0x306d, 0x306e,
-    0x306f, 0x3072, 0x3075, 0x3078, 0x307b, 0x307e, 0x307f, 0x3080, 0x3081,
-    0x3082, 0x3084, 0x3086, 0x3088, 0x3089, 0x308a, 0x308b, 0x308c, 0x308d,
-    0x308f, 0x3093,
-  ];
+  hiraganaLookup: Record<string, string> = {
```
melink14, line 294: Since this is a big table it might be worth putting it in it's own file (with other similar constants)

extension/data.ts
```diff
@@ -296,93 +287,308 @@ class RcxDict {
...
-    // note: katakana vu is never converted to hiragana
-
-    p = 0;
-    reason = '';
-    for (i = 0; i < word.length; ++i) {
-      u = v = word.charCodeAt(i);
-
-      // Skip Zero-width non-joiner used in Google Docs between every
-      // character.
-      if (u === 8204) {
```
melink14, line 346: I couldn't tell but do you still have the logic which skips zero-width non-joiner? I saw it was in your table but didn't see it referenced your new code.

extension/data.ts
```diff
@@ -296,93 +287,308 @@ class RcxDict {
...
     word: string,
     doNames: boolean,
     max?: number
   ): DictEntryData | null {
     let i;
-    let u;
-    let v;
-    let reason: string;
-    let p;
     const trueLen = [0];
```
melink14, line 582: It doesn't like you update this anymore, is that okay?

melink14 (changes requested):
Didn't look at the tests yet due to time constraints but mostly nits with some high level questions on if the new code is functionally equivilant to the old code.

Also, it seems there are still some unexpected screenshot diffs! (title translations aren't continuing past the first word?)

melink14 (commented):
Another thought: I wonder if we should just use wanakana library to do all this stuff: https://www.npmjs.com/package/wanakana

Seems to have what we need and is small.

melink14:
> I did wan't to ask you about that. I have to update my executablePath to make sure pupetteer works correctly. Do you think I should just add it to my global .gitignore or something?

I forgot to answer this question but this can probably be updated to fallback to an environment variable or command line option.

I might also test if it's even required anymore (since there should be some logic to find chrome in a normal way) but I think I was debugging some problems in WSL which made me add it originally.

tora-pan:
So I went ahead and commented out the `executablePath:` portion and the tests still run so I think I've found our solution. 👍

extension/data.ts
```diff
@@ -244,46 +243,38 @@ class RcxDict {
     return null;
   }
 
-  deinflect(word: string) {
+  deinflect(word: string): Deinflection[] {
     const r = [];
```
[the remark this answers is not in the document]
tora-pan: Noted and taken care of.

extension/data.ts
```diff
@@ -244,46 +243,38 @@ class RcxDict {
...
     const have: { [key: string]: number } = {};
-    let o;
 
-    o = { word: word, type: 0xff, reason: '' } as Deinflection;
-    r.push(o);
+    r.push({ word: word, type: 0xff, reason: '' });
```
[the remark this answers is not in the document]
tora-pan: Yes, this is exactly what I did.

extension/data.ts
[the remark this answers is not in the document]
tora-pan: Love it! 💯

extension/data.ts
tora-pan: Yeah, I have it in the lookup table which maps to the same character (essentially skipping / not converting it)

extension/data.ts
tora-pan: Totally agree 💯

extension/data.ts
tora-pan: Yeah close. Maybe the `.all` is python? Here is what I refactored it to. LMK what you think.

```isKana = Array.from(word).every((char) => this.isKana(char.charCodeAt(0)));```

File: extension/data.ts [hunk 11]
```diff
@@ -244,46 +243,38 @@ class RcxDict {
...
       const type = r[i].type;
-
-      for (j = 0; j < this.difRules.length; ++j) {
-        const g = this.difRules[j];
-        if (g.fromLength <= wordLen) {
-          const end = word.substr(-g.fromLength);
-          for (k = 0; k < g.rules.length; ++k) {
-            const rule = g.rules[k];
+      for (let j = 0; j < this.difRules.length; j++) {
+        const currentDifRule = this.difRules[j];
```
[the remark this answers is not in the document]
tora-pan: Yeah sorry, it was a typo. I was thinking in my head "current deinflection rule" I think what came out ended up being "current deflection rule" and then typo'd into currentDifRule. haha

extension/data.ts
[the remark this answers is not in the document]
tora-pan: 💯

extension/data.ts
```diff
@@ -296,93 +287,308 @@ class RcxDict {
...
-    0x3092, 0x3041, 0x3043, 0x3045, 0x3047, 0x3049, 0x3083, 0x3085, 0x3087,
-    0x3063, 0x30fc, 0x3042, 0x3044, 0x3046, 0x3048, 0x304a, 0x304b, 0x304d,
-    0x304f, 0x3051, 0x3053, 0x3055, 0x3057, 0x3059, 0x305b, 0x305d, 0x305f,
-    0x3061, 0x3064, 0x3066, 0x3068, 0x306a, 0x306b, 0x306c, 0x306d, 0x306e,
-    0x306f, 0x3072, 0x3075, 0x3078, 0x307b, 0x307e, 0x307f, 0x3080, 0x3081,
-    0x3082, 0x3084, 0x3086, 0x3088, 0x3089, 0x308a, 0x308b, 0x308c, 0x308d,
-    0x308f, 0x3093,
-  ];
+  hiraganaLookup: Record<string, string> = {
+    ぁ: 'ぁ',
```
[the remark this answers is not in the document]
tora-pan: exactly. Rather than checking if it is full-width katakana or half-width katakana and then converting, I figured it would be cleaner/easier to check if it is kana and then use the lookup map regardless. 

I added the hiragana characters which I don't believe were in the original conversion charts.

extension/data.ts
```diff
@@ -296,93 +287,308 @@ class RcxDict {
...
+    ﾖ: 'よ',
+    ﾗ: 'ら',
+    ﾘ: 'り',
+    ﾙ: 'る',
+    ﾚ: 'れ',
+    ﾛ: 'ろ',
+    ﾜ: 'わ',
+    ｦ: 'を',
+    ﾝ: 'ん',
+    ｳﾞ: 'ゔ',
```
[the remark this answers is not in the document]
tora-pan: You know what, I think you are right. I did a quick check of the original and even though the conversion chart has `0x30f4` the range that it checks is `<=  0x30f3` which is one before.

I'll remove it.

tora-pan:
@melink14 all tests are passing now. Fixed a few logic issues. Going to do a bit more cleaning up of variable names and typing in the morning tomorrow.

🥂

extension/data.ts
tora-pan: Are we referring to trueLen? It is updated on line 713.

extension/data.ts
melink14: But iiuc, the important part of of this check is not that we don't convert it but that it's removed from the final word after conversion. The way this code works is that at the end of each loop the new converted character is appended to the `reason` variable and then when all characters are converted `word` is reset to the new `reason` value.

if we're just converting it directly then the final word we use to do the search will still have this useless character in it, right?

I think part of the problem is that it's shoehorned in with the conversion logic; it might be better to have a separate part of the code which strips out unneeded characters and then group them up in something like a `normalize` method...

tora-pan: You are 100% right. This could cause issues with the lengths. I just did a sample test and event though it's invisible, it gets counted with `.length`

extension/data.ts
```diff
@@ -296,93 +287,308 @@ class RcxDict {
...
+    ぃ: 'ぃ',
+    い: 'い',
+    ぅ: 'ぅ',
+    う: 'う',
+    ぇ: 'ぇ',
+    え: 'え',
+    ぉ: 'ぉ',
+    お: 'お',
+    か: 'か',
+    が: 'が',
```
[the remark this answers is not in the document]
tora-pan: I think the will match because in the case of half-width characters (the ones where this matters), we check the trailing character to see if it is a voice/semi-voiced mark. if so, we replace the previously added string with a new lookup combining the 2 characters together.

like this:
```
const key = isSemiVoiced
        ? char + nextChar
        : isVoiced
        ? char + nextChar
        : char;
      const hiragana = kanaToHiraganaNormalizationMap[key];

```

extension/data.ts
```diff
@@ -199,190 +188,152 @@ class RcxDict {
...
-    let j;
-    let k;
-
-    i = 0;
-    do {
-      word = r[i].word;
-      const wordLen = word.length;
-      const type = r[i].type;
+    for (let i = 0; i < possibleDeinflections.length; i++) {
+      const currentDeinflection = possibleDeinflections[i].word;
```
melink14, line 237: it seems odd to call this `currentDeinflection` since it's a sub field of `possibleDeinflections`

Either maybe the `word` subfield should be updated to be a better name for what it represents or we should use something like `currentDeinflectionWord` here or simlar.

extension/data.ts
```diff
@@ -199,190 +188,152 @@ class RcxDict {
...
+  }
 
-  cv: number[] = [
-    0x30f4, 0xff74, 0xff75, 0x304c, 0x304e, 0x3050, 0x3052, 0x3054, 0x3056,
-    0x3058, 0x305a, 0x305c, 0x305e, 0x3060, 0x3062, 0x3065, 0x3067, 0x3069,
-    0xff85, 0xff86, 0xff87, 0xff88, 0xff89, 0x3070, 0x3073, 0x3076, 0x3079,
-    0x307c,
-  ];
-  cs: number[] = [0x3071, 0x3074, 0x3077, 0x307a, 0x307d];
+  convertToHiragana(str: string): string {
```
melink14, line 294: Maybe we can make `str` more semantic. `kanaCharacter` or `character`.

Actually, is this the whole word, in that case `word` is fine as well.

Since this method actually expects special kinds of strings it's good to use something that conveys that.

extension/data.ts
```diff
@@ -199,190 +188,152 @@ class RcxDict {
...
+    let result = '';
+    for (let i = 0; i < str.length; i++) {
+      const char = str.charAt(i);
+      const nextChar = i + 1 <= str.length - 1 ? str.charAt(i + 1) : null;
```
melink14, line 298: I don't think you need this boundary checking. I did a quick experiment and `str.charAt` returns `''` if you go past the length of the string.

extension/data.ts
```diff
@@ -199,190 +188,152 @@ class RcxDict {
...
+      const nextCharCode = nextChar?.charCodeAt(0);
```
melink14, line 299: BTW, is there any reason to use character codes to do this check or can we just do direct string comparison?

melink14, line 299: IIUC, next char only matters if the current character is a half width katakana. Should we add a check for that to make it clear.

I guess if these punctation only show up in this case then we can call this a simplification but it might make understanding the flow easier in the future if we're more explicit. (that is, make the code closely mirror our goals rather than relying on implicit properties of our inputs)

extension/data.ts
```diff
@@ -199,190 +188,152 @@ class RcxDict {
...
+      const isSemiVoiced = nextCharCode === PUNCTUATION.SEMI_VOICED_MARK;
+      const isVoiced = nextCharCode === PUNCTUATION.VOICED_MARK;
+
+      const key = isSemiVoiced
```
melink14, line 303: I think this can be simplified a bit maybe:

`const key = isSemiVoiced || isVoiced ? char + nextChar : char`

extension/data.ts
```diff
@@ -199,190 +188,152 @@ class RcxDict {
...
+        ? char + nextChar
+        : isVoiced
+        ? char + nextChar
+        : char;
+      const hiragana = kanaToHiraganaNormalizationMap[key];
+      result += hiragana !== undefined ? hiragana : char;
```
melink14, line 309: In this case, since we're already handling undefined can we make the lookup table smaller so that it only contains special characters we have to remap? That would make it more obvious which cases are special!

extension/data.ts
melink14: That's the reading of `trueLen` though. I'm talking about the updating of it to be something other than `[0]`.

Is 
```
// need to keep real length because of the half-width semi/voiced
// conversion
trueLen[reason.length] = i + 1;
```

not needed anymore or are we representing it a new way?

extension/types/data-types.ts
```diff
@@ -0,0 +1,45 @@
+export type DictData = {
```
melink14, line 1: I think we talked about it a bit on discord but after seeing it; I think this types based cateogorization is not as useful as a semantic categorization (were files are grouped based on their semantic purpose rather than on a more general classification)

It's also fairly unclear what constitutes a type for some of the files here.

Finally, we should be careful when adding new exports as that creates public interface contracts that we'll need to be careful in the future.

I think if we want to add some grouping then we could group all the dictionary handling code together (maybe renaming data.ts (probably spliting it in the future since it contains fairly distinct functiaonality around loading and lookups now))

(we could also punt this til later since it's not so complicated yet and it's good to keep code moving in their own PRs (otherwise git doesn't detect it and gives you a bad diff as you can see here)

extension/types/hiraganaLookupMap.ts
```diff
@@ -0,0 +1,247 @@
+export const kanaToHiraganaNormalizationMap: Record<string, string> = {
```
melink14, line 1: I wouldn't consider this a type at all since it contains implementation details. But it's definitely data related so to me it's easier to group stuff when thinking about the product rather than the language features.

extension/types/hiraganaLookupMap.ts
```diff
@@ -0,0 +1,247 @@
...
+  ぁ: 'ぁ',
```
melink14, line 2: As I mentioned elsewhere, what about just removing the 1:1 mappings here since we already handle that at the call site.

Right now it's a lot to sift through to see which things are actually being changed.

extension/types/unicode-constants.ts
```diff
@@ -0,0 +1,17 @@
+export enum KANA {
+  HIRAGANA_START = 0x3041,
+  HIRAGANA_END = 0x309f,
+  KATAKANA_START = 0x30a0,
+  KATAKANA_END = 0x30ff,
+  HW_KATAKANA_START = 0xff66,
+  HW_KATAKANA_END = 0xff9d,
+}
+
+export enum PUNCTUATION {
```
melink14, line 10: Since these are used to check equality is there any reason to use their code points rather than strings?

extension/types/unicode-constants.ts
```diff
@@ -0,0 +1,17 @@
...
+  VOICED_MARK = 0xff9e,
+  SEMI_VOICED_MARK = 0xff9f,
+  NON_JOINER = 0x200c,
+  // These two are only used in full-width and may not be needed
```
melink14, line 14: if not needed we should leave out so people aren't distracted. :)

extension/types/unicode-constants.ts
```diff
@@ -0,0 +1,17 @@
+export enum KANA {
```
melink14, line 1: you could probably combine this with the lookup map if you called it something like kana_data.ts or kana_info.ts or something like that. Or since it's not all kana just character_info.ts.

Naming things is hard but there's a balance to strike on how fine grained groupings should be.

melink14:
I see the images are still there but that might be due to the presubmit not re-running to update them. (Since it doesn't re-run when github action pushes a commit)

I'll close and reopen the pull request to get it to re-run. (I should add a comment or label command for it...)

melink14 (changes requested):

extension/data.ts
```diff
@@ -42,50 +42,36 @@
...
 };
-type DictEntryData = typeof defaultDictEntryData;
 
 interface Deinflection {
   word: string;
   type: number;
   reason: string;
 }
 
-interface DeinflectionRule {
```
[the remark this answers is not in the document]
tora-pan: I see. That makes sense. I'll move them back in and remove the types file.

extension/types/unicode-constants.ts
tora-pan: We can change them to the actual strings, however, I thought for consistency we could keep them all as code points. I guess it is just a bit of OCD on my end. 

Having all code points and then one :
`ZERO_WIDTH_JOINER = '‌',` <- The character is actually inside of the quotes.

Also I thought this would be slightly confusing to look at.
```
console.log("～".charCodeAt(0));
console.log("〜".charCodeAt(0));
```
from first glance they appear to be the same but they are actually different.
```
65374 or 0xff5e
12316 or 0x301c
```

If you think otherwise, I can just swap it back to having the characters themselves.

extension/types/unicode-constants.ts
tora-pan: These have been removed.

extension/types/unicode-constants.ts
tora-pan: I like that. `character_info.ts`

extension/types/hiraganaLookupMap.ts
tora-pan: Types folder has been removed, constants have been added to the lookup map and the file has been renamed to `character_info.ts`

extension/data.ts
tora-pan: yes, this is the whole word. Do you want to name it `inflectedWord` or `japaneseWord` or `conjugatedWord` instead? Or does `word` suffice for this?

extension/data.ts
tora-pan: Good catch again. This was an overlook on my part. It very well should be that. 

I'm thinking `curDeinflectionWord` to cut down a tad bit on the length. Starting to look more and more like Java.

Sound good?

extension/data.ts
```diff
@@ -199,190 +188,152 @@ class RcxDict {
...
-    let i;
-    let j;
-    let k;
-
-    i = 0;
-    do {
-      word = r[i].word;
-      const wordLen = word.length;
-      const type = r[i].type;
+    for (let i = 0; i < possibleDeinflections.length; i++) {
```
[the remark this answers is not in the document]
tora-pan: Sounds good to me. Replacing the original `Deinflect` method

tora-pan:
> Another thought: I wonder if we should just use wanakana library to do all this stuff: https://www.npmjs.com/package/wanakana
> 
> Seems to have what we need and is small.
> 
> edit: maybe doesn't handle half width...

This hasn't been updated in years but looks like it does what we need it to do?
https://github.com/yomotsu/japanese-string-utils
Maybe we could either use it or get "hints" from it?

extension/data.ts
tora-pan: I think this bug should be fixed now. The refactor changes removed the need for `trueLen`. After that, I updated the `maxLen` to what the actual selection should be an local tests seems to be working as expected. 🙏

melink14:
> This hasn't been updated in years but looks like it does what we need it to do?
[yomotsu/japanese-string-utils](https://github.com/yomotsu/japanese-string-utils)
Maybe we could either use it or get "hints" from it?

Might be worth trying though if we found problems I guess we'd need to fork it.

extension/types/unicode-constants.ts
melink14: I guess both ways can make it hard to tell at a glance what the character is but at least with the unicode style, it's unambiguous in that you know how to look it up.

I think I was just thinking it seemed weird to convert a string to a code point when we controlled the lookup table.

One thing I'd recommend (if we're not already doing this) is add a comment with the string character next to/above the unicode symbols in case someone wants to copy it quickly. Might require some clever thinking to make it obvious in the case of invisible characters. :)

extension/data.ts
melink14: It's more clear if we store the characters as unicode as discussed in the other thread.

extension/data.ts
melink14: maybe `mixedKanaWord` or something that conveys the idea of what kind of string we would need to use this function on. (Naming things is an art not science)

extension/data.ts
```diff
@@ -277,112 +295,80 @@ class RcxDict {
...
+  }
 
-  cv: number[] = [
-    0x30f4, 0xff74, 0xff75, 0x304c, 0x304e, 0x3050, 0x3052, 0x3054, 0x3056,
-    0x3058, 0x305a, 0x305c, 0x305e, 0x3060, 0x3062, 0x3065, 0x3067, 0x3069,
-    0xff85, 0xff86, 0xff87, 0xff88, 0xff89, 0x3070, 0x3073, 0x3076, 0x3079,
-    0x307c,
-  ];
-  cs: number[] = [0x3071, 0x3074, 0x3077, 0x307a, 0x307d];
+  convertToHiragana(str: string): string {
```
melink14, line 332: I might have said this elsewhere but definititely recommend a jsdoc for this method and maybe a tweak to the name since it also has a truncation function now (which we added back for consistency)

melink14 (changes requested):
I still haven't looked at the tests yet but I can feel the implementation is pretty close probably.

extension/data.ts
tora-pan: What do you mean by `truncation` portion? If you mean the stripping of the ZWJ and Tilde, etc... then that happens before it is sent to the `convertToHiragana` method.

Here is what I currently have using TS-Doc:

Screenshot 2023-03-08 at 7 06 05 AM

What do you think?

extension/data.ts
tora-pan: Awesome catch. You are correct and removing the boundary check doesn't cause any other issues. 
```
const nextChar = kanaWord.charAt(i + 1);
const nextCharCode = nextChar?.charCodeAt(0);
```
`nextCharCode` ends up being `NaN` but I don't think there is anything to worry about as it's only being used to check for `isVoiced` or `isSemiVoiced`.

extension/data.ts
tora-pan: Awesome I didn't know I could do that. 
Thank you.

extension/data.ts
```diff
@@ -277,112 +295,96 @@ class RcxDict {
...
+      const currentChar = kanaWord.charAt(i);
+      const currentCharCode = currentChar.charCodeAt(0);
+      let isSemiVoiced = false;
+      let isVoiced = false;
+      const isHalfWidthKatakana =
+        currentCharCode >= KANA.HW_KATAKANA_START &&
+        currentCharCode <= KANA.HW_KATAKANA_END;
+      let key = '';
+      if (currentCharCode < 0x3000) {
+        break;
```
melink14, line 352: This is the truncation I meant; with this line the string will be truncated at the first not kana/kanji character.

extension/data.ts
melink14: It's at line 352; I left a comment there for easy reference.

The TS doc is solid but I'll leave some pedantic comments (mostly conforming to best practices developed in my day job but that have served me well)

extension/character_info.ts
```diff
@@ -0,0 +1,186 @@
+export const kanaToHiraganaNormalizationMap: Record<string, string> = {
```
melink14, line 1: Should this be constant case? (and readonly I guess)

extension/data.ts
```diff
@@ -277,112 +295,96 @@ class RcxDict {
...
-  ];
+  isKana(charCode: number): boolean {
+    return (
+      (charCode >= KANA.HIRAGANA_START && charCode <= KANA.HIRAGANA_END) ||
+      (charCode >= KANA.KATAKANA_START && charCode <= KANA.KATAKANA_END) ||
+      (charCode >= KANA.HW_KATAKANA_START && charCode <= KANA.HW_KATAKANA_END)
+    );
+  }
+  /**
+   * Returns the input string converted into hiragana. If any characters are not
```
melink14, line 332: nit: It's good to reference the inputs expliictly in the summary statement so that you don't need to use an `@param`. The `@` markers can be useful for additional information that doesn't flow naturally in the summary statement but there's no need to repeat information there that can be read once at the beginning!

So in this case something like:
```
Returns the given `kanaWord` with katakana and half width characters converted to full-width hiragana. ヴ is not converted. If a non-kana characters is found in `kanaWord`, that and the following characters are omitted from the returned conversion.
```

extension/data.ts
```diff
@@ -277,112 +295,96 @@ class RcxDict {
...
+   * found in the [NormalizationMap](./character_info.ts) then the character
```
melink14, line 333: This information is all true but it's more of an implementation detail vs a spec. Ideally, the doc string should be about the goal of the method and needn't go into detail about how we accomplished that. Basically, anything that could be changed without updating a test, shouldn't go in the doc string.

melink14 (changes requested):

melink14 (approved):

melink14:
Ignoring codecov/patch since the line was already uncovered but the refactoring shifted the lines some.

PR #1446, opened 2023-03-09
fix: Add logic to return early if hover returns empty string

### LINKED ISSUE #1417: Hovering over ruby text pops up orangatang kanji
There's a code that returns empty string when hovering over ruby text. This empty string is passed through the logic and ends up returning the orangatang kanji at unicode 0000.

We should add a check to just return early in this case (and maybe reconsider hovering over ruby text?)

tora-pan:
Add logic to return early if hover returns empty string.

Fixes #1417


extension/rikaichan.ts
```diff
@@ -263,7 +263,10 @@ class RcxMain {
 
     const m = this.showMode;
     let e: DictEntryData | null = null;
-
+    if (text === '') {
```
melink14, line 266: Add a test for this branch (as you probably saw from the coverage report. :) )

melink14 (changes requested):
We talked about also updating the content scripts as well (such that we'd expect to never hit this early return.

If we're treating this as the fix for the bug, maybe we should open another issue for tracking cleaning up the client side (so we don't send these search requests).

Another thought I had was whether this should be here or in the background.ts message handler. It's sometimes said that request validation should happen at system entry points (essentially, means you can safely assume valid data everywhere else)

I don't have a strong opinion but curious to hear your thoughts about it.

extension/rikaichan.ts
tora-pan: Yeah, Let me do some digging and thinking for your above question and also handle the tests for this if this is the approach we end up going with 👍

tora-pan: @melink14 , I moved the quick catch to the `background.ts`. I also added tests to cover 99% of the file. 

Had trouble covering this line. [Here](https://github.com/melink14/rikaikun/blob/d2a52a41a8619c9282e9a995b25cb5f1a2e52ec9/extension/background.ts#L38)

I tried a few different approaches and even tried using a 3rd party package `chai-as-promised` to assist with async testing to no avail. 

I'll leave it to you or circle back later. 👍

melink14: That sounds good. You should revert the changes to thsi file then I guess!

Also, the additional tests for expanding coverage are of course good but I wouldn't include it in the same PR as one fixing a specific bug. It makes it hard for a reviewer and future reader to know which tests were directly related to the fix and which were just updating code quality in general.

Is it possible to do a quick PR split and keep things very focused?

tora-pan: You got it.

extension/rikaichan.ts
```diff
@@ -263,7 +263,6 @@ class RcxMain {
 
     const m = this.showMode;
     let e: DictEntryData | null = null;
-
```
melink14, line 266: Looks like there was a whitespace change left in.

We should definitely talk about dev environments since some flows will make this more obvious.

extension/test/background_test.ts
```diff
@@ -65,26 +87,76 @@ describe('background.ts', function () {
...
+
+      // eslint-disable-next-line @typescript-eslint/await-thenable
+      await chrome.runtime.onMessage.addListener.callArgWith(
+        0,
+        request,
+        sender,
+        response
+      );
+
+      expect(searchStub.calledWith(request.text, request.dictOption)).to.be
```
melink14, line 116: This is a bit non idiomatic for an expectation. I would expect:

`expect(searchStub).toHaveBeenCalledWith(request.text, request.dictOption)`

Though what I said above I would say:

`expect(searchStub).toHaveBeenCalledWithMatch(request.text, sinon.match.any)`

(exact syntax my differ)

melink14, line 116: I wrote the other comment for information but in generally verify intermediate state isn't necessary if you are testing the inputs and outputs.

In this case, you can drop this and you can know it's working by just testing the response is called with the right arguments.

If you want to ensure the right arguments are passed you can update the stub rather than using a verification. Usually, though keeping stubs loose is good for behavior testing. 

(this is pretty subtle so it's something to think about each test and build up some intuition)

extension/test/background_test.ts
```diff
@@ -65,26 +87,76 @@ describe('background.ts', function () {
...
+        .true;
+
+      expect(response.called).to.be.true;
```
melink14, line 119: I would update this to check that it was called with the right output since it's the key behavior of the code under test.

extension/test/background_test.ts
```diff
@@ -65,26 +87,76 @@ describe('background.ts', function () {
...
+      // eslint-disable-next-line @typescript-eslint/await-thenable
+      await chrome.runtime.onMessage.addListener.callArgWith(
+        0,
+        request,
+        sender,
+        response
+      );
+
+      expect(searchStub).to.not.be.called;
+      expect(response.called).to.be.false;
```
melink14, line 138: I think this one isn't so important (and probably why it's not in the test description).

In fact, I wonder if we should return something to response but it seems like it works fine if it never gets a response but if it got some special response then it's probably fine since the goal is to return early.

package.json
```diff
@@ -27,6 +27,7 @@
     "test": "wtr \"extension/test/*_test.ts\"",
     "test:small": "wtr --files \"extension/test/background_test.ts\" \"extension/test/data_test.ts\" \"extension/test/docs-annotate-canvas_test.ts\" \"extension/test/rikaicontent_test.ts\"",
     "test:browserstack": "npm run test -- --browserstack",
+    "test:tiny": "wtr --files \"extension/test/background_test.ts\"",
```
melink14, line 30: probably don't need this all the time; If a person wants to run a single test they can use `.only` or just run wtr on the command line right?

melink14 (changes requested):

extension/test/background_test.ts
tora-pan: So in this case, I would have to test that the response callback  is being called with a class method that passes in the arguments?`rcxMain.search(request.text, request.dictOption)` ?

Can we just change the test to: `expect(searchStub.calledWith(request.text, sinon.match.any)).to.be.true;`

and then drop:
`expect(response.called).to.be.true` and call it a day?

extension/test/background_test.ts
tora-pan: Honestly, I agree, however it is also nice to have 100% branch coverage isn't it?

package.json
tora-pan: This one leaked through. Sorry.

extension/test/background_test.ts
```diff
@@ -65,26 +87,76 @@ describe('background.ts', function () {
...
-    { type: type },
-    { tab: { id: tabId } },
-    responseCallback
-  );
-  return;
-}
+  describe('xsearch', function () {
+    it('should return search results', async function () {
+      const searchStub = sinon.stub().returns(['to eat', 'verb']);
+      rcxMain.search = searchStub;
```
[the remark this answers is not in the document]
tora-pan: When I do it inline and change the test to `expect(rcxMain.search)...`  Typescript ESLint complains:
```
Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead.
```

When switching it to an arrow function to fix this error, ESLint complains that mocha tests should not use arrow functions.

extension/test/background_test.ts
```diff
@@ -65,26 +87,76 @@ describe('background.ts', function () {
         { config: rcxMain.config }
       );
     });
+
+    it('should throw typeError if tab is undefined', async function () {
+      // }).to.throw(TypeError, 'sender.tab is always defined here.');
```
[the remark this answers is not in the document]
melink14: it's not removed yet. :)

I recommend only marking conversations as resolved after you've pushed since otherwise reviewers can't easily confirm the solution.

extension/test/background_test.ts
melink14: > So in this case, I would have to test that the response callback is being called with a class method that passes in the arguments? `rcxMain.search(request.text, request.dictOption)` ?

This doesn't sound right. `response` is called with the return value of `RcxMain.search` which you have stubbed at the top of the test. So you just need to confirm it's called with the values you passed to `sinon.stub`

Or I misunderstood! I don't think any class methods are involved...

> and call it a day?

Actually, the opposite is usually what we want from a readability/maintainability point of view.

drop `expect(searchStub.calledWith...)`

And add: `expect(response).toBeCalledWith(...)`

the amount of code and structure is basically the same but the points of focus are the different:

With the way I'm suggesting, you're passing through the inputs and confirming the outputs. This is contrast to verifying some internal method call which is hidden from the user.

extension/test/background_test.ts
melink14: This shouldn't affect test coverage should it? Was it reported in the coverage report?

IIUC, both of these are in the same statement so by testing that searchStub is not called it's the same as confirming that response is not called.

The goal here is to focus the user's attention on the goal of the underlying code as documented in the test.

If we did care about `response`, my first thought is to make a new test like `should not call response callback if text input is empty` or similar.

melink14 (changes requested):

extension/test/background_test.ts
```diff
@@ -65,26 +87,76 @@ describe('background.ts', function () {
...
+        type: 'xsearch',
+        text: 'た',
+        dictOption: '2',
+      };
+      const sender = { tab: { id: 0 } };
+      const response = sinon.spy();
+
+      // eslint-disable-next-line @typescript-eslint/await-thenable
+      await chrome.runtime.onMessage.addListener.callArgWith(
+        0,
```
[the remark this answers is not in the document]
tora-pan: Thank you for the recommendation!

tora-pan:
@melink14 I'm not sure of the correct procedure for this but, moving this to a new branch / pr to simplify things.

https://github.com/melink14/rikaikun/pull/1458

extension/test/background_test.ts
```diff
@@ -65,26 +87,76 @@ describe('background.ts', function () {
...
-  return;
-}
+  describe('xsearch', function () {
+    it('should return search results', async function () {
+      const searchStub = sinon.stub().returns(['to eat', 'verb']);
+      rcxMain.search = searchStub;
+      const request = {
+        type: 'xsearch',
+        text: 'た',
+        dictOption: '2',
```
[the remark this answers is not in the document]
tora-pan: Not sure I follow. This is testing the `case: 'xsearch'` so if we don't pass it into the request, it won't be caught be the switch statement right?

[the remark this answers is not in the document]
melink14: Sorry this comment is attached to line 103 so I meant `dictOption` when I said type. I think I forgot there was an actual key called `type` and just meant in general since this refers to the dictionary type.

extension/background.ts
```diff
@@ -33,26 +33,24 @@ chrome.runtime.onMessage.addListener(async (request, sender, response) => {
   const rcxMain = await rcxMainPromise;
   switch (request.type) {
     case 'enable?':
-      console.log('enable?');
```
melink14, line 36: can you revert these deletions?

melink14 (changes requested):
You marked some comments as resolved but they still seem to be here. Did you push?

melink14:
There is still https://github.com/melink14/rikaikun/pull/1458/files which is similar to this but loses the comments. It'd be nice if you just copied the conents of https://github.com/melink14/rikaikun/pull/1458/files to this branch and then closed the other PR. (since they're the same except this one has history)

extension/background.ts
tora-pan: Completed with latest push.

melink14 (approved):

PR #1447, opened 2023-03-09
build(deps): Remove stylelint-config-prettier dep and references in stylelint config

### LINKED ISSUE #1402: Remove stylelint-config-prettier
As of version 15, stylelint has removed formatting lint rules and thus the default should be compatible with prettier.

ref: https://github.com/prettier/stylelint-config-prettier/releases/tag/v9.0.5

tora-pan:
stylelint removed all their formatting rules so now the default package is compatible with prettier with no additional plugins.

Fixes #1402


melink14 (approved):

melink14:
I fixed a typo in the PR title and then edited it a bit to remove some of the overlap in the title and the body. (So the title is more what but the body is more why + extra details)

PR #1449, opened 2023-03-12
test: Expand test coverage of background.ts to 100% coverage

tora-pan:


melink14 (changes requested):
Looks like it still contains the changes from the other PR. :(

Also, I'm not sure why the title was changed as it was since you already have a PR which is about fixing the bug!

As I said in the chat, you can try again or let me know if you want to work through it together!

tora-pan:
@melink14 Do I need to add these tests changes to their own branch? Will that clean things up?

melink14:
> @melink14 Do I need to add these tests changes to their own branch? Will that clean things up?

Yes, in this one I would just make a new branch off of main and then copy just the test file into it. (we can also wait til we land the first change if you want to reuse the helper methods you made)

melink14 (approved):

PR #1458, opened 2023-03-14
fix: Return early if search string is empty

tora-pan:
Added catch for empty strings being passed in to return early.

Updated `background.test.ts` helper function `sendMessageToBackground` to take in optional `text` argument.

Fixes: #1417


tora-pan:
Closing as changes have been captured on [this pr](https://github.com/melink14/rikaikun/pull/1446)

PR #1464, opened 2023-03-19
refactor: Use template literals instead of string concatenation when building popup HTML

tora-pan:
The inline variable replacement of template literals allows us to remain the desired HTML shape of the string.

Renamed variables for better readability.

### DIFF
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -668,7 +668,7 @@ class RcxDict {
       return '';
     }
 
-    const htmlToRender = [];
+    const htmlToRender: string[] = [];
 
     // kanji view
     if (entry.kanji) {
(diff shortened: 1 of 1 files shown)


melink14 (commented):
This seems fine overall but will need to take a block of time to study it since it produced a non-trivial diff.

Since several of the branches don't have tests (not your fault) we might want to add those first so we can be confident in the refactor. (I can do it as well if you don't have time or desire)

Finally, I agree the variable renaming is good but I would probably have done that in it's own PR first since it leads to cleaner diffs and easier to review PR. That said, it's only a minor inconvenience in this case so it's not worth struggling with git to split it! It's something to consider in the future though!

I also updated the title and body a little bit to have the title focus on exactly what was done and the body give extra context for why (and thanks for separately mentioning renames! that's very important when combining changes!). It was very close already though so only needed minor tweaks ('need' is a strong word here)

Issue #1280, opened 2022-11-17
Turn on strict typescript flags in tsconfig.json


melink14:
When we can we should turn on:
- noImplicitAny
- noUnusedLocals
- noUncheckedIndexedAccess

Found from testing which flags were required in typescript playground for rikaicontent.ts to compile.

tora-pan:
This could be quite the undertaking. 

Do you think it would be possible to do this incrementally?

My thoughts would be to do something like:
`tsc --p tsconfig.strict.json && tsc --p tsconfig.nonstrict.json`

```
// strict.tsconfig.json
{
  "compilerOptions": {
    "strict": true
  },
  "include": [
    "a.ts"
  ]
}
```
and
```
// tsconfig.nonStrict.json
{
  "compilerOptions": {
    "strict": false
  },
  "include": [
    "b.ts",
    "c.ts"
  ]
}
```

Not sure if this would work or not but if so, it would allow us to do this file by file.

Thoughts?

melink14:
To be clear, strict mode is already enabled for all files with exceptions for the flags as defined in the first post! (ref https://www.typescriptlang.org/tsconfig#strict)

That said, I think it's a great idea to allow flexibility if some files are easier to migrate than others. We could actually make a file for each flag or make a file for all three stricter flags depending on how easy/hard it was.

melink14:
I was a bit off. `noImplicitAny` is already turned on since strict mode is on and as far as I can tell, we're not manually setting it to false.

The other two are separate from `strict` mode but just seem like good ideas I think.

I think this issue title was a bit misleading since it's goal is stricture TS flags but by using the word 'strict' it made it easy to confuse with that flag in particular.

(or I'm misunderstanding something about the current TS config)

tora-pan:
Sorry, I shouldn't have used the word "strict" what I meant was just to have different tsconfig files with different levels of strictness, "flags" turned on.

I think I remember previously when I turned them on, there were hundreds of warnings/errors so I wanted to find a way to decrease the amount of changes for each PR.

melink14:
Did you try only the flags listed here or other flags too.

unused locals should be fairly easily to auto-fix (since they're unused!) unless some missing type information makes some of the alerts false positives.

Unchecked index access is trickier since it requires accurate types and there are some places in rikaikun which were written in a pretty type free manner...

Either way, the first step would be to finalize the set of flags we want to attempt and then I think breaking down like you said is a good idea!

PR #1601, opened 2023-06-04
build(tsconfig): Add noUnusedLocals to tsconfig.json

bryanjenningz:
- Add noUnusedLocals property mentioned in this issue: https://github.com/melink14/rikaikun/issues/1280
- Remove unused `dictCount` property detected by added flag.

Other useful tsconfig flags can be handled in subsequent pull requests

### DIFF
--- a/extension/rikaicontent.ts
+++ b/extension/rikaicontent.ts
@@ -65,7 +65,6 @@ type Rikaichan = {
 };
 
 class RcxContent {
-  private dictCount = 3;
   private altView = 0;
 
   private sameDict = 0;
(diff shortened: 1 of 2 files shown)


extension/rikaicontent.ts
```diff
@@ -65,7 +65,6 @@ type Rikaichan = {
 };
 
 class RcxContent {
-  private dictCount = 3;
```
bryanjenningz, line 68: After adding "noUnusedLocals" to the tsconfig.json file, it told me that this dictCount property was not used, so I removed it. I double-checked to make sure that this dictCount property wasn't being used anywhere by searching the code (https://github.com/search?q=repo%3Amelink14%2Frikaikun%20dictCount&type=code).

melink14:
Thanks Bryan for the PR. It looks good to me. I made some tweaks to the PR title/description (nits only) but otherwise it looks good. (I'm moving away from the chore category as being too catch all)

melink14 (approved):

Issue #1453, opened 2023-03-13
ＤＶＤ-ＲＯＭ isn't recognized


melink14:
I think it's due to the code which returns early when unicode < 0x3000 is detected in a string.

We might want to look at the data and see what other punctuation is being missed.

Expectation:
ＤＶＤ-ＲＯＭ [ディーブイディーロム] /(n) (comp) DVD-ROM/

Issue #190, opened 2020-07-26
Handle ～ properly


melink14:
We usually ignore ～ when it's in the middle of a word since it's just style but by itself it has a meaning which we can keep.
rikaichamp did this in https://github.com/birtles/rikaichamp/commit/fa1c7f77f826c0342d7aa82da3a230050f67ad2c

It's important to do since right now the update-db.js script will add an empty index for this character since it's in the dictionary and that needs to be manually updated.

Alternatively, we could fix the word dictionary parsing to skip tilde as well and then ideally it wouldn't show up anywhere but that seems a bit worse.

PR #1607, opened 2023-06-06
fix(search): Handle word searches with char codes below 0x3000

bryanjenningz:
Words like ＤＶＤ-ＲＯＭ were not matching due to the hyphen (represented by a char code below 0x3000) in the middle. In order to include, entries like ＋α, logic was removed from rikaicontent.ts that skipped lookups when the letter the mouse hovered over was not a normal Japanese char code. (This included other various ranges which didn't show additional impacts as of this PR.)

These changes are thought to have been added for performance reasons which are no longer relevant but if some problem does crop up we can solve it more narrowly and with better documentation at that time.

<details>
  <summary>See word entries affected by this change (122 unique words)</summary>

```ts
[
  "〇×",
  "〇×テスト",
  "＋α",
  "○",
  "○×",
  "○×テスト",
  "○○",
  "◎",
  "△",
  "※印",
  "２×４工法",
  "２進-５進",
  "４°",
  "ＣＤ-ＤＡ",
  "ＣＤ-Ｒ",
  "ＣＤ-ＲＯＭ",
  "ＣＯＶＩＤ-１９",
  "ＤＶＤ-Ｒ",
  "ＤＶＤ-ＲＡＭ",
  "ＤＶＤ-ＲＯＭ",
  "ＨＡＮＡ-ＢＩ",
  "Ｈα線",
  "Ｊ-ＰＯＰ",
  "Ｋ-１",
  "Ｋ-ＰＯＰ",
  "Ｋ-Ｔ境界",
  "ＲＦ-ＩＤ",
  "ＵＴＦ-８",
  "Ｗｉ-Ｆｉ",
  "ウラン-２３８",
  "プラスα",
  "マイクロRNA",
  "マッチョ♀ウーマン",
  "Α",
  "Β",
  "Γ",
  "Γ分布",
  "Δ",
  "Ε",
  "Ζ",
  "Η",
  "Θ",
  "Θ関数",
  "Ι",
  "Κ",
  "Λ",
  "Μ",
  "Ν",
  "Ξ",
  "Ο",
  "Π",
  "Ρ",
  "Σ",
  "Τ",
  "Υ",
  "Φ",
  "Χ",
  "Ψ",
  "Ω",
  "α",
  "α-ｈｅｌｉｘ",
  "α-ヘリックス",
  "α-メチルトリプタミン",
  "α-リノレン酸",
  "αらせん",
  "αメチルトリプタミン",
  "αリノレン酸",
  "α化",
  "α化米",
  "α線",
  "α波",
  "α分類学",
  "α米",
  "α崩壊",
  "α螺旋",
  "α粒子",
  "β",
  "β-カロチン",
  "β-カロテン",
  "βカロチン",
  "βカロテン",
  "βテスト",
  "βバージョン",
  "β線",
  "β波",
  "β版",
  "β崩壊",
  "β粒子",
  "γ",
  "γ-オリザノール",
  "γアミノ酪酸",
  "γリノレン酸",
  "γ線",
  "γ分布",
  "γ崩壊",
  "δ",
  "ε",
  "ζ",
  "η",
  "θ",
  "θ理論",
  "ι",
  "κ",
  "λ",
  "μ",
  "μ粒子",
  "ν",
  "ξ",
  "ο",
  "π",
  "π中間子",
  "π電子",
  "ρ",
  "σ",
  "σ電子",
  "τ",
  "τ粒子",
  "υ",
  "φ",
  "χ",
  "ψ",
  "ω"
]
```
This list of words was collected by running this code:

```ts
const uniqueWords = [
  ...new Set(
    rcxDict.wordDict
      .trim()
      .split("\n")
      .map((line) => line.split(" ")[0])
  ),
];
const wordsWithCharCodesBelow0x3000 = uniqueWords.filter((word) =>
  [...word].some((char) => char.charCodeAt(0) < 0x3000)
);
console.log(
  "wordsWithCharCodesBelow0x3000",
  JSON.stringify(wordsWithCharCodesBelow0x3000, null, 2)
);
```
</details>

I wrote snapshot tests to make sure none of the words without character codes below 0x3000 were affected by this change. I didn't include those tests in this pull request because they take a while to run (since there are over 200,000 entries in the dictionary) and they make this pull request bigger. I can add those tests in a separate pull request if we think that's a good idea so we can verify diffs of word searches for all words in the dictionary. This would give us more confidence whenever we change the wordSearch function because it shows all the diffs for each search result in the dictionary and it would allow us to drastically simplify the wordSearch function in the future with confidence that we aren't breaking of the previous behavior.

I also wrote code to do a check to make sure all words in the dictionary are searchable and found there is only a 1 unsearchable word in the dictionary now:

```ts
const uniqueWords = [
  ...new Set(
    rcxDict.wordDict
      .trim()
      .split("\n")
      .map((line) => line.split(" ")[0])
  ),
];
const unsearchableWords = uniqueWords.filter(
  (word) =>
    !rcxDict
      .wordSearch(word, false)
      ?.data.find((data) => data.entry.startsWith(word + " "))
);
console.log("unsearchableWords", unsearchableWords); // Logs ['～']
```

So this means after this pull request gets merged, the only unsearchable word entry in the dictionary is "～ [にょろ] /(n) tilde/wave dash/".

This word is a single character with a character code of 65374, which is 0xff5e in hexadecimal (which is above 0x3000, which explains why this pull request doesn't handle this word). After searching through the code for this number 0xff5e, it looks like there's some code that handles this character code that needs to be modified slightly to account for this word in the dictionary. That can be handled in a subsequent pull request to keep this pull request short and simple.

Fixes: https://github.com/melink14/rikaikun/issues/1453


extension/data.ts
```diff
@@ -86,6 +86,17 @@ interface DeinflectionRuleGroup {
...
+// prettier-ignore
+const CHAR_CODES_BELOW_0x3000_USED_IN_WORDS_IN_DICTIONARY = [
+  '×', 'α', '○', '◎', '△', '※', '-', '°', 'R',
+  'N', 'A', '♀', 'Α', 'Β', 'Γ', 'Δ', 'Ε', 'Ζ',
+  'Η', 'Θ', 'Ι', 'Κ', 'Λ', 'Μ', 'Ν', 'Ξ', 'Ο',
+  'Π', 'Ρ', 'Σ', 'Τ', 'Υ', 'Φ', 'Χ', 'Ψ', 'Ω',
+  'β', 'γ', 'δ', 'ε', 'ζ', 'η', 'θ', 'ι', 'κ',
+  'λ', 'μ', 'ν', 'ξ', 'ο', 'π', 'ρ', 'σ', 'τ',
+  'υ', 'φ', 'χ', 'ψ', 'ω'
+].map(char => char.charCodeAt(0));
```
bryanjenningz, line 98: Let me know if I should change this code so that it has the character codes already computed rather than calling Array.prototype.map here.

I chose to leave it like this so that the characters are human-readable, but I can also change it to just show the character codes if that's preferred.

extension/data.ts
```diff
@@ -86,6 +86,17 @@ interface DeinflectionRuleGroup {
   rules: DeinflectionRule[];
 }
 
+// prettier-ignore
+const CHAR_CODES_BELOW_0x3000_USED_IN_WORDS_IN_DICTIONARY = [
```
bryanjenningz, line 90: Let me know if I should name this to something else. I couldn't think of a good name, so I wrote a very long constant variable.

Also let me know if I should move this constant so that it's in the `RcxDict` class as a static constant so that it follows the convention of the other constants.

extension/test/data_test.ts
```diff
@@ -62,5 +62,11 @@ describe('data.ts', function () {
         rcxDict.wordSearch('ぼんです', /* doNames= */ false)?.data
       ).to.not.include.something.like({ entry: /^凡打 .*/ });
     });
+
+    it('should return results for words with punctuation', function () {
+      expect(
+        rcxDict.wordSearch('ＤＶＤ-ＲＯＭ', /* doNames= */ false)?.data
+      ).to.include.something.like({ entry: /^ＤＶＤ-ＲＯＭ .*/ });
+    });
```
bryanjenningz, line 70: I'm not sure if I should add more tests. I only wrote one test which covers the case mentioned in https://github.com/melink14/rikaikun/issues/1453.

extension/data.ts
bryanjenningz: I got this list of character codes below 0x3000 by running this code:

```ts
const uniqueWords = [
  ...new Set(
    rcxDict.wordDict
      .trim()
      .split("\n")
      .map((line) => line.split(" ")[0])
  ),
];
const charactersBelow0x3000InDictionaryWords = [
  ...new Set(
    uniqueWords.flatMap((word) =>
      [...word].filter((char) => char.charCodeAt(0) < 0x3000)
    )
  ),
];
console.log(
  "charactersBelow0x3000InDictionaryWords",
  charactersBelow0x3000InDictionaryWords
);
```

melink14, line 90: Using this reduces the scope of the change but also means the problem could crop up again if a new symbol is added....

As you were working on this did you come across any problems if we just remove the check altogether? It might have been added for performance reasons only...

extension/test/data_test.ts
melink14: Judging from code coverage, I think we should also add a test for a word that contains a character less than 3000 that is not in the CONSTANT such that both halves of the conditional are testsed.

extension/data.ts
melink14: This seems reasonable to me.

melink14 (changes requested):
Just a quick review  before work; I'll try to look again afterwards! (It looks reasonable though!)

melink14:
https://github.com/melink14/rikaikun/issues/190 is the issue for handling ～

As far other tests, I think we don't need to test every single word; it should be enough to test a representative word and be confident that similar words behave the same. I'd be interested more in the doubts you had while developing since as you say the goal of tests is to allow making changes without worrying!

bryanjenningz:
> As far other tests, I think we don't need to test every single word; it should be enough to test a representative word and be confident that similar words behave the same.

Yeah, I agree. I'll just add a few words in different categories of words and that's probably enough. I didn't see any words that were handled incorrectly so this should be sufficient.

> I'd be interested more in the doubts you had while developing since as you say the goal of tests is to allow making changes without worrying!

I didn't have any specific doubts while developing. I'm just not familiar with the code in wordSearch so I just wrote some tests to see the diffs before and after my change to be 100% sure the wordSearch output is changed how I expected it to change.

extension/data.ts
bryanjenningz: Removing the if-check for character codes under 0x3000 is much more simple and it doesn't appear to introduce any performance problems when I test it locally. Thanks!

extension/data.ts
```diff
@@ -348,10 +348,6 @@ class RcxDict {
         continue;
       }
 
-      if (u <= 0x3000) {
-        break;
-      }
-
```
bryanjenningz, line 354: It doesn't cause any performance problems when the if-check here is removed.

extension/rikaicontent.ts
```diff
@@ -848,20 +848,6 @@ class RcxContent {
       }
     }
 
-    //
-    if (
```
bryanjenningz, line 852: The wordSearch function now returns results for all words in the dictionary except "～", but I noticed that some words (e.g. "＋α") still get blocked at this if-check before wordSearch is ever called. I'm guessing it's there for performance (similar to the 0x3000 char code check in wordSearch). I removed this if-check and it doesn't appear to have any performance problems and it fixed this problem for words like "＋α".

melink14: This seems good to me. I can see the goal was to maybe not waste lookups if the first character was seemingly not Japanese looking but that excludes various words in the dictionary and is probably not needed anymore.

If a problem comes up later we can solve it more narrowly then with proper documentation.

melink14:
I changed the PR body a bit to make it read better (imo) once it gets merged and becomes the commit message body. Just for your information, I had two goals:
 
 - Remove information that was already in the PR title.
 - Make the initial paragraph give additional context around why the changes were done (including the final rikaicontent change)

An astute reader would be able to figure that stuff out but since code sleuthing months/years later is hard, I tend to try to make it as easy as possible for my future self. :)

extension/rikaicontent.ts
melink14: Though it might be good to add an rikaicontent (or e2e) test for a word like ＋α so that it doesn't get broken in the future. What do you think?

(Something like https://github.com/melink14/rikaikun/blob/7023f15c59bca8573f2120b05a3dd5ac033d64a6/extension/test/rikaicontent_test.ts#LL120C20-L120C20 for example but with different text in the test data.)

bryanjenningz:
> I changed the PR body a bit to make it read better (imo) once it gets merged and becomes the commit message body. Just for your information, I had two goals:
> 
>     * Remove information that was already in the PR title.
> 
>     * Make the initial paragraph give additional context around why the changes were done (including the final rikaicontent change)
> 
> 
> An astute reader would be able to figure that stuff out but since code sleuthing months/years later is hard, I tend to try to make it as easy as possible for my future self. :)

Thanks so much! I appreciate that. The PR has a nice initial paragraph now. I'll be sure to add more context like this in future PRs so that it's easier to understand for my future self too.

extension/rikaicontent.ts
bryanjenningz: I think that's a great idea. I added an e2e test for non-Japanese characters triggering xsearch similar to https://github.com/melink14/rikaikun/blob/7023f15c59bca8573f2120b05a3dd5ac033d64a6/extension/test/rikaicontent_test.ts#LL120C20-L120C20.

melink14 (approved):

melink14:
@all-contributors Add @bryanjenningz for code.

PR #1613, opened 2023-06-12
refactor: Delete unused functions fileReadArray and makeHtmlForRuby

bryanjenningz:
This pull request removes 2 unused functions in the data.ts file: `fileReadArray` and `makeHtmlForRuby`.

Feel free to close this pull request if we're planning on using these functions in the future.

If you use code search, you can see they're unused - so they can be safely removed:
- https://github.com/search?q=repo%3Amelink14%2Frikaikun%20fileReadArray&type=code
- https://github.com/search?q=repo%3Amelink14%2Frikaikun+makeHtmlForRuby&type=code

### DIFF
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -155,17 +155,6 @@ class RcxDict {
     return req.responseText;
   }
 
-  fileReadArray(name: string) {
-    const a = this.fileRead(name).split('\n');
-    // Is this just in case there is blank shit in the file.  It was written
-    // by Jon though.
-    // I suppose this is more robust
-    while (a.length > 0 && a[a.length - 1].length === 0) {
-      a.pop();
-    }
-    return a;
-  }
-
   loadNames() {
     if (this.nameDict && this.nameIndex) {
       return;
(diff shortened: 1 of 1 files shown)


melink14 (approved):
Thanks!

My opinion is that dead or unused code should never be left in. We can look at the history if we think we want it again.

If these are the last ones we may be able to add a compile or lint check to block future ones.

bryanjenningz:
It's worth mentioning that TypeScript was unable to detect that these functions were unused because they were public functions. If we set these functions to private functions, then TypeScript gives an error saying that they're unused.

melink14:
That makes sense; as a Chrome Extension, we know about all possible usages but Typescript doesn't have that context. (Also we set `isolatedModules` in tsconfig which means it only compiles one file at a time.)

This is probably a niche problem, mostly exacerbated by the non idiomatic typescript that resulted from migrating it all at once form legacy code. (That said, in a world where we break up the files into smaller modules, it would be nice to have this capability)

Thanks again!

PR #1620, opened 2023-06-13
test(makeText): Add `makeText` tests 

bryanjenningz:
I'm adding some tests to `makeText` because I plan on simplifying it in subsequent pull requests.


extension/test/data_test.ts
```diff
@@ -69,4 +69,60 @@ describe('data.ts', function () {
       ).to.include.something.like({ entry: /^ＤＶＤ-ＲＯＭ .*/ });
     });
   });
+
+  describe('makeText', function () {
+    it('should return the text for a word dictionary entry', function () {
+      const entry = rcxDict.wordSearch('あ', /* doNames= */ false);
+      if (!entry) {
```
melink14, line 76: I don't think we need this check since `makeText` accepts a nullable entry. If entry is null then the test will fail due to the assertions not matching. 

Having a single 'exit' point in a test makes it easier to reason and avoids mixing different types of test failures (and having a clear Arrange Act Assert structure)

(It's not unreasonable to want to confirm the setup data is correct but it won't be too hard to trace this type of failure with the existing asserts. Tests are mainly living documentation and will be read many more times than they're debugged due to failing; thus I prefer to optimize for readability in this case)

extension/test/data_test.ts
```diff
@@ -69,4 +69,60 @@ describe('data.ts', function () {
...
+        throw new Error(
+          `rcxDict.wordSearch('あ', /* doNames= */ false) returns null`
+        );
+      }
+      expect(rcxDict.makeText(entry, /* max= */ 1)).to.equal(
```
melink14, line 81:
```suggestion

      expect(rcxDict.makeText(entry, /* max= */ 1)).to.equal(
```

It's convenient to explicitly call out the Arrange Act Assert sections with a single new line between each part. (In this case Act and Assert are in the same statement which is fine imo since vertical space is precious)

https://github.com/goldbergyoni/javascript-testing-best-practices#-%EF%B8%8F-12-structure-tests-by-the-aaa-pattern

extension/test/data_test.ts
```diff
@@ -69,4 +69,60 @@ describe('data.ts', function () {
       ).to.include.something.like({ entry: /^ＤＶＤ-ＲＯＭ .*/ });
     });
   });
+
+  describe('makeText', function () {
+    it('should return the text for a word dictionary entry', function () {
```
melink14, line 74: `text` is a bit vague since it's hard for a reader to know the test is correct just from reading the test description and test body.

In a perfect world, I would write this test such that different parts of of the text output were tested independently so if a bug in one part of the text generation failed I could immediately see exactly which part failed due to which test failed.

In this case, the it serves as a pretty good regression test to help ensure no backsliding happens but isn't as good documentation.

Feel free to not change the format but keep in mind the larger philosphy going forward (more important for new code)

extension/test/data_test.ts
```diff
@@ -69,4 +69,60 @@ describe('data.ts', function () {
...
+      const entry = rcxDict.wordSearch('あ', /* doNames= */ false);
```
melink14, line 75: I think ideally we'd construct the entry our selves instead of relying on `wordSearch` this would allow us to explicitly see what parts of the entry are important to the output. Right now, the reader needs to understand how `wordSearch` works to know if the test is correct and why.

It also makes the test  less precise since a bug in `wordSearch` will cause the test to fail.

That said, this is still a clear improvement so feel free to keep like this and just remember the overall idea.

extension/test/data_test.ts
```diff
@@ -69,4 +69,60 @@ describe('data.ts', function () {
...
+      if (!entry) {
+        throw new Error(
+          `rcxDict.wordSearch('あ', /* doNames= */ false) returns null`
+        );
+      }
+      expect(rcxDict.makeText(entry, /* max= */ 1)).to.equal(
+        'あ\t(int) (1) ah; oh; (int) (2) hey!\n',
+        `word: 'あ', doNames: false, max: 1`
+      );
+      expect(rcxDict.makeText(entry, /* max= */ 2)).to.equal(
```
melink14, line 85: Prefer to test a single behavior per test. In this case, you're testing two different `max` values ensuring it works.

Instead, I would make this second assert its own  `it` block with a name including something like "includes entries up to max" or similar.

You can share the entry by using a second level `describe` which scopes tests that are about word search.

melink14 (changes requested):
I think this seems pretty good. It's not quite what I would do for new code but it's perfectly suited to the purpose of getting ready for future modifications and avoiding regressions.

That said, are you really planning on making changes to `makeText`? Does that mean you want to impact the copy to clipboard functionality vs the regular popup?

bryanjenningz:
> That said, are you really planning on making changes to makeText? Does that mean you want to impact the copy to clipboard functionality vs the regular popup?

I'm not going to be making any changes to the functionality of makeText (so it should work exactly how it works now) - I just noticed that makeText and makeHtml both use similar parsing logic so it might make sense to do the parsing in one place so that there's less repetition. So it shouldn't impact the copy to clipboard functionality or the regular popup.

extension/test/data_test.ts
bryanjenningz: Good catch. I completely agree.

extension/test/data_test.ts
bryanjenningz: Yeah, it might help with readability to have spaces in between the arrange part, the action part, and the assertion part. Thanks for the link. I'll try follow this style when writing tests.

extension/test/data_test.ts
bryanjenningz: Thanks for the tip. I agree. The tests should be organized so that when one fails, you have a good idea of what failed just by reading the test description. I'll improve this.

extension/test/data_test.ts
bryanjenningz: Yeah, I agree - good point. I want these makeText tests to only fail if the makeText function fails to adhere to its specification, but the way I wrote this test makes it possible for wordSearch to cause this test to fail, which makes the tests harder to understand when they fail.

extension/test/data_test.ts
bryanjenningz: Good idea! I'll do that.

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
       ).to.include.something.like({ entry: /^ＤＶＤ-ＲＯＭ .*/ });
     });
   });
+
+  describe('makeText', function () {
+    describe('with a word dict entry', function () {
+      // wordDictEntry was generated with: rcxDict.wordSearch('あ', /* doNames= */ false)
```
bryanjenningz, line 75: I'm not sure if this comment is helpful. If it's not helpful, I can remove it.

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
...
+          /* max= */ 1000
+        );
+
+        expect(allDataEntriesAsText).to.equal(
+          'あ\t(int) (1) ah; oh; (int) (2) hey!\nア\t(int) (1) ah; oh; (int) (2) hey!\n'
+        );
+      });
+    });
+
+    describe('with a name dict entry', function () {
```
bryanjenningz, line 128: The makeText function doesn't treat entries from the name dict differently than entries from the word dict, so it might not be necessary to test an entry from the name dict. I can delete these name dict entry tests in this describe block if we think these name dict tests aren't necessary.

Maybe instead of testing a name dict entry, I can add more kanji dict tests to get full test coverage (for example, I could test a kanji dict entry where the "bushumei" property isn't an empty string).

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
...
+          { entry: 'あ /(int) (1) ah/oh/(int) (2) hey!/', reason: undefined },
+          { entry: 'ア /(int) (1) ah/oh/(int) (2) hey!/', reason: undefined },
+        ],
+        hasMore: false,
+        title: '',
+        index: 0,
+        matchLen: 1,
+      };
+
+      it('when max is 1, returns 1 data entry as text', function () {
```
bryanjenningz, line 94: I tried to improve the wording here and it's a little bit better than before, but let me know if you have a better way to word this so it's more clear about what's being tested.

extension/test/data_test.ts
melink14: I think removing it makes sense to me. The structure of the entry object is independent and over time it's possible that the output of `wordSearch` and these tests will diverge making the comment a lie.

extension/test/data_test.ts
bryanjenningz: Before I just wrote "text" which was vague, but I think "data entry as text" is a little bit better at least (but it's maybe still not ideal).

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
       ).to.include.something.like({ entry: /^ＤＶＤ-ＲＯＭ .*/ });
     });
   });
+
+  describe('makeText', function () {
+    describe('with a word dict entry', function () {
+      // wordDictEntry was generated with: rcxDict.wordSearch('あ', /* doNames= */ false)
+      const wordDictEntry = {
```
melink14, line 76: This is a definite improvement!

One thing I notice though is there are a lot of fields which seem to not impact the test results but take up a lot of space. These make the test harder to quickly parse for the important inputs.

A reasonable goal is to hide all parts of the input which wouldn't change the outcome of the test. Whatever's left is critical and stands out (and is probably referenced in the test description at least indirectly).

There are at least two options for doing this:

1. If the function/type allows for undefined/null fields then you can just exclude the fields from the object.
2. If there are fields which are required but the value doesn't impact this specific test you can create a default object (or a factory method) that you use as the base for creating your test inputs. (Careful to never actually rely on the default values in assertions!)

    This is pretty easy to do with the spread operator I believe!

In this case, we definitely need `hasNames: false` and the `data` but I'm not sure about stuff like `index` and `matchLen`. It's possible they are derived from `data` so also important. That connection will be hard to show unfortunately so a comment might be the best we can do.

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
...
+        kanji: '',
+        onkun: '',
+        nanori: '',
+        bushumei: '',
+        misc: {},
+        eigo: '',
+        hasNames: false,
+        data: [
+          { entry: 'あ /(int) (1) ah/oh/(int) (2) hey!/', reason: undefined },
```
melink14, line 85: Similar to above, can we just exclude reason or does it have to be undefined to get the same answer. (maybe in that case add a note to one of the descriptors)

melink14, line 85: It's reasonable to use a real entry for this test data but my preference is to use fake data which draws the users focus to the parts of the input data which are critical.

In this case, the primary transformation is about taking the `/` separated parts and formatting them differently. In that case, you could do something like.

`entry: '<head1> /<sense 1.1>/<sense 1.2>/'`
`entry: '<head2> /<sense2.1>/<sense 2.2>/`

This really highlights that the actual text doesn't matter at all but the separator does. It also shows that the space in the head is removed in the formatted output. (I used `<>` since it's commonly used to show placeholders but something else (or nothing) might be more readable)

(of course, I didn't test this so if for example the `makeText` has some constraints on the input then we'd need to tweak the test data but that would again highlight exactly which parts of the input matter)

extension/test/data_test.ts
melink14: This is reasonable except that at least the input is changed with `hasNames` now being `true`. Does that value not really effect the `makeText` logic at all? In that case, it makese sense but maybe we can make the description of the other test more broad.

This case does show what happens when there's a kanji plus pronunciation which might be useful to test independtly of whether it's a name or not...

The decision to add more kanji tests is an independent one in my opinion. Definitely, testing all the fields would be expected if the code was new, but I'll leave it you to decide how much backfill you want to do. :)

Given the current kanji test has almost all fields entered, you could easily just add a non empty string to the bushumei property... (In general, more fine grained tests might be better but better to fix in the future as new code gets introduced)

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
...
+          { entry: '亜 [あ] /(f) A/', reason: undefined },
+          { entry: '阿 [あ] /(s) A/', reason: undefined },
+        ],
+        hasMore: false,
+        title: '',
+        index: 0,
+        matchLen: 1,
+      };
+
+      it('when max is 1, returns 1 data entry as text', function () {
```
melink14, line 148: Relevent to the comment about needing these tests; even if we thought a name test was useful, we probably don't need to test as many cases as before. A single test focusing on the fact the that it's a name is enough given the behavior of `max` isn't different at all. (This won't matter if we delete all of them of course)

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
...
+          kanjiDictEntry,
+          /* max= */ 1
+        );
+
+        expect(singleKanjiDataEntryAsText).to.equal(
+          '両\nboth; old Japanese coin; counter for carriages (e.g., in a train); two\nリョウ、 てる、 ふたつ\n名乗り\tもろ\nHalpern\t3518\nHeisig 5th Edition\t1168 both\nHeisig 6th Edition\t1252 both\nHenshall\t411\nKanji Learners Dictionary\t2191\nKanji Learners Dictionary 2nd Edition\t2949\nNelson\t34\nNew Nelson\t23\nPinYin\tliang3\nSkip Pattern\t4-6-1\nTuttle Kanji & Kana\t200\nTuttle Kanji Dictionary\t0a6.11\nUnicode\t4E21\n'
+        );
+      });
+
+      it('when max is 1000, returns a single kanji data entry as text', function () {
```
melink14, line 222: I would tweak this to make it more obvious that kanji entries ignore max essentially. Maybe:

`even with large max, returns a single kanji data entry as text`

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
...
+          /* max= */ 1000
+        );
+
+        expect(allDataEntriesAsText).to.equal('亜\tあ\t(f) A\n阿\tあ\t(s) A\n');
+      });
+    });
+
+    describe('with a kanji dict entry', function () {
+      // kanjiDictEntry was generated with: rcxDict.kanjiSearch('両')
+      const kanjiDictEntry = {
```
melink14, line 178: the same comments about input data apply here as above though the kanji flow uses more fields. (This actually makes me think that this entry type could be split into one for kanji and one for text...)

extension/test/data_test.ts
```diff
@@ -69,4 +69,166 @@ describe('data.ts', function () {
...
+        kanji: '両',
+        onkun: 'リョウ、 てる、 ふたつ',
+        nanori: 'もろ',
+        bushumei: '',
+        misc: {
```
melink14, line 183: For new code, I would probably have a single test for each misc field but for a regression test it's not required.

melink14 (changes requested):
Changes look good. I think the last really relevant bit is hiding irrelevant details of entry in the tests!

extension/test/data_test.ts
bryanjenningz: I was thinking the same thing about having separate types for kanji and words since the fields used in the kanji entries and the word entries seem to be completely different. We can separate the types in a later pull request to keep this pull request small and focused on tests.

extension/test/data_test.ts
bryanjenningz: Yeah, that sounds better. I'll do that.

extension/test/data_test.ts
bryanjenningz: Yeah I agree, using fake data that draws attention to what should be focused on will make the tests even easier to understand, so I'll do that.

bryanjenningz: We can't exclude the reason field because the reason field's type is written ["reason: string | undefined"](https://github.com/melink14/rikaikun/blob/main/extension/data.ts#L55), but if we change the type to "reason?: string" then we will be able to exclude the "reason: undefined" part in the tests. Let me know if I can change the type to "reason?: string" or if I should just leave the type alone for now in this pull request.

melink14: Interesting, I thought excluding a field was the same as the field being undefined in typescript. There are other options (mentioned in the other comment) for avoiding irrelevant properties but this one is short so I would leave as is.

I think in the long term this property being optional is correct. I'm not sure why an explicit `undefined` type would ever by preferred off the top of my head.

bryanjenningz: If you use `{ reason?: string }`, then `{ reason: undefined }` and `{}` are both allowed.
If you use `{ reason: string | undefined }`, then `{ reason: undefined }` is allowed but `{}` is not allowed.

I checked in the TypeScript playground [here](https://www.typescriptlang.org/play?noUnusedParameters=true&ssl=10&ssc=1&pln=1&pc=1#code/C4TwDgpgBA8mwEsD2A7AhgGwEoQMZICcATKAXigG8oCI0BnVAfgC4o7gCEUBzKAXwDcAKCH4U7KEnjJ02PISIBGVnESpMOfMTKVqtBilYBXFEQgAzLhBKDRqCVLWzNCgEwrp6uVpLkKtoVBIKABVUwsrIhdtPz16VFZ2Th4oAB8oEzNLFGt+YTtxYAzw7OtopVYwrMjynSoaeMNi6pybYTEJTIjW8vdQkpr5GMpBKAB6Magg6AgCAkIhIA).

bryanjenningz: The only benefit I could think of for using `{ reason: string | undefined }` would be if you have some code that depends on the property always being there, for example maybe if you have code that checks the number of entries in the object for some reason (maybe something like `Object.keys(obj).length` which is expected to always be the same), but this seems unlikely, so it's probably fine to use `{ reason?: string }`.

extension/test/data_test.ts
bryanjenningz: I ended up removing these name dict entry tests that I originally added because the `hasNames` property doesn't affect the return value so the name dict entry tests seemed unnecessary - the only property that matters for non-kanji-dict entries is the `entry` property in the `data` array, so I updated the tests to make that more clear.

I also ended up adding more tests for the kanji dict entries to get more test coverage.

extension/test/data_test.ts
bryanjenningz: I ended up deleting these tests because `makeText` doesn't treat name dict entries any differently than word dict entries.

extension/test/data_test.ts
```diff
@@ -69,4 +69,209 @@ describe('data.ts', function () {
       ).to.include.something.like({ entry: /^ＤＶＤ-ＲＯＭ .*/ });
     });
   });
+
+  describe('makeText', function () {
+    describe('with a non-kanji dict entry', function () {
+      const unimportantProperties = {
```
melink14, line 75: Two things:
1.  Usually, we'd define this outside the test as a module local constant so that it's not taking up space while reading the test. The idea is that if the defaults are truly unimportant then they don't need to be seen while reading.

2.  Related, if you move it outside you'd probably want one that worked for all your tests. In that sense, instead of specifying the exact unimportant properties, you would just have a `DEFAULT_DICT_ENTRY` or similar and use that anywhere you needed to customize.

extension/test/data_test.ts
```diff
@@ -69,4 +69,209 @@ describe('data.ts', function () {
       ).to.include.something.like({ entry: /^ＤＶＤ-ＲＯＭ .*/ });
     });
   });
+
+  describe('makeText', function () {
+    describe('with a non-kanji dict entry', function () {
```
melink14, line 74: having an empty kanji is important but so is having present `data` property. Maybe there's a way to have both.

`with a dict entry consisting of empty kanji but present data properties` or similar...

extension/test/data_test.ts
```diff
@@ -69,4 +69,209 @@ describe('data.ts', function () {
...
+          kanjiDictEntry,
+          /* max= */ 1000
+        );
+
+        expect(singleKanjiDataEntryAsText).to.equal(
+          '<kanji-entry>\n<eigo-entry-1>; <eigo-entry-2>; <eigo-entry-3>\n<onkun-entry-1>、 <onkun-entry-2>、 <onkun-entry-3>\n名乗り\t<nanori-entry>\n部首名\t<bushumei-entry>\nHalpern\t<H-misc-entry>\nHeisig 5th Edition\t<L-misc-entry>\nHeisig 6th Edition\t<DN-misc-entry>\nHenshall\t<E-misc-entry>\nKanji Learners Dictionary\t<DK-misc-entry>\nKanji Learners Dictionary 2nd Edition\t<DL-misc-entry>\nNelson\t<N-misc-entry>\nNew Nelson\t<V-misc-entry>\nPinYin\t<Y-misc-entry>\nSkip Pattern\t<P-misc-entry>\nTuttle Kanji & Kana\t<IN-misc-entry>\nTuttle Kanji Dictionary\t<I-misc-entry>\nUnicode\t<U-misc-entry>\n'
+        );
+      });
+
+      describe('empty fields', function () {
```
melink14, line 198: maybe `with empty fields` for consistency; it would probably make sense either way in the final output...

extension/test/data_test.ts
```diff
@@ -69,4 +69,209 @@ describe('data.ts', function () {
...
+        it('uses a hyphen after kanji field when eigo field is empty', function () {
+          const emptyEigoKanjiDictEntry = { ...kanjiDictEntry, eigo: '' };
+
+          const hyphenAsEigoEntryText = rcxDict.makeText(
+            emptyEigoKanjiDictEntry,
+            /* max= */ 1
+          );
+
+          expect(hyphenAsEigoEntryText).to.equal(
+            '<kanji-entry>\n-\n<onkun-entry-1>、 <onkun-entry-2>、 <onkun-entry-3>\n名乗り\t<nanori-entry>\n部首名\t<bushumei-entry>\nHalpern\t<H-misc-entry>\nHeisig 5th Edition\t<L-misc-entry>\nHeisig 6th Edition\t<DN-misc-entry>\nHenshall\t<E-misc-entry>\nKanji Learners Dictionary\t<DK-misc-entry>\nKanji Learners Dictionary 2nd Edition\t<DL-misc-entry>\nNelson\t<N-misc-entry>\nNew Nelson\t<V-misc-entry>\nPinYin\t<Y-misc-entry>\nSkip Pattern\t<P-misc-entry>\nTuttle Kanji & Kana\t<IN-misc-entry>\nTuttle Kanji Dictionary\t<I-misc-entry>\nUnicode\t<U-misc-entry>\n'
```
melink14, line 208: I wonder if we can use regex matching to ignore the parts of the text we don't care about... (or change the input to have fewer set fields)

Focusing the assertion makes it much easier to match the goal in the name to what's actually happening.

extension/test/data_test.ts
```diff
@@ -69,4 +69,209 @@ describe('data.ts', function () {
...
+
+        it('omits the nanori field after onkun field when nanori field is empty', function () {
+          const emptyNanoriKanjiDictEntry = { ...kanjiDictEntry, nanori: '' };
+
+          const noNanoriFieldKanjiEntryText = rcxDict.makeText(
+            emptyNanoriKanjiDictEntry,
+            /* max= */ 1
+          );
+
+          expect(noNanoriFieldKanjiEntryText).to.equal(
```
melink14, line 220: For `omits` is some klind of `does.not.contain` better; though that wouldn't be as specific as including the relationship to onkun...

melink14 (changes requested):
This looks pretty good to me; especially the extra tests for kanji entries. I think the remaining requests are pretty small so it shouldn't be long before the merge!

extension/test/data_test.ts
bryanjenningz: Yeah that's a good point, I'll add that so the test is more explicit about the requirements.

It seems like right now the `kanji` property is used to determine whether a dict entry is a kanji dict entry or a non-kanji/word dict entry. That's kind of confusing because there are a bunch of properties that the kanji dict entry ignores and properties that the non-kanji/word dict entry ignores. So maybe in the future we might want to refactor the code so that these types are more explicit and easier to understand. It might help with readability and it might make impossible types impossible (e.g. it should be impossible for a word dict entry to have a non-empty kanji property). Maybe something like this:
```ts
type DictEntry = KanjiDictEntry | WordDictEntry;
type KanjiDictEntry = {
  kind: 'kanji';
  // properties only used by kanji dict entry
};
type WordDictEntry = {
  kind: 'word';
  // properties only used by word dict entry
};
```
Then in the code which wants to deal with a dict entry, we could maybe get exhaustive type checking which is more explicit, something like this:
```ts
switch (dictEntry.kind) {
  case 'kanji':
    return // do things with kanji entry
  case 'word':
    return // do things with word entry
}
```
Not sure if this is a good idea, but just brainstorming ways to make the code more explicit and easier to understand in the future.

melink14: Something like that seems good I agree.

Using Discriinated Unions like you describe seems reasonable! Another idea would be to consider if any functions actually need to operate on both kinds of dict entries. For example, `makeTextForKanji` and `makeTextForWord` as separate functions could be better depending how exactly they're called. 

In general, branching early is better than later (makes code easy to reason about) but it really depends on how much shared code there is!

Definitely worth thinking about. There's a ton of legacy code that I inherited so no need to hold back. :D

melink14 (approved):
Thanks for the changes! I think the intent and readability are near perfect now! Of course, even the first draft was quite good so I appreciate your particpating to make it even better.

PR #1623, opened 2023-06-15
refactor(makeText): Rename variables, move variables to local scope

bryanjenningz:
The goal of this pull request is to give variables in `makeText` more descriptive names so that the code is easier to understand. Some variables were also put into a more local scope so that they're easier to follow.

This pull request does not change the functionality of `makeText` - the copy-to-clipboard feature works exactly how it did before this pull request.

I tried to make this pull request have a small code diff so that it's easier to review. I plan on making more changes to `makeText` in subsequent pull requests which will make it even easier to understand.

Changes I plan on making in subsequent pull requests:
- Restructure the `kanjiInfoLabelList` array so that it's an array of objects - so its type might change to something like `{ kanjiInfoCode: string; kanjiInfoName: string }[]` so that it's easier to understand.
- Write a dict entry parsing function which is well-tested and unifies the parsing logic used in `makeText`, `makeHtml`, and the update-db.ts file.
- Change the `DictEntryData` type to represent the valid states of the application better - something like `type DictEntryData = KanjiDictEntry | WordDictEntry;`

### DIFF
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -940,7 +940,6 @@ class RcxDict {
   }
 
   makeText(entry: DictEntryData | null, max: number): string {
-    let i;
     let j;
     let t;
 
@@ -962,7 +961,7 @@ class RcxDict {
         b.push('\u90E8\u9996\u540D\t' + entry.bushumei + '\n');
       }
 
-      for (i = 0; i < this.kanjiInfoLabelList.length; i += 2) {
+      for (let i = 0; i < this.kanjiInfoLabelList.length; i += 2) {
         const e = this.kanjiInfoLabelList[i];
         j = entry.misc[e];
         b.push(
@@ -976,7 +975,7 @@ class RcxDict {
       if (max > entry.data.length) {
         max = entry.data.length;
       }
-      for (i = 0; i < max; ++i) {
+      for (let i = 0; i < max; ++i) {
         const e = entry.data[i].entry.match(
           /^(.+?)\s+(?:\[(.*?)\])?\s*\/(.+)\//
         );
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -961,8 +961,8 @@ class RcxDict {
       }
 
       for (let i = 0; i < this.kanjiInfoLabelList.length; i += 2) {
-        const e = this.kanjiInfoLabelList[i];
-        const j = entry.misc[e];
+        const kanjiInfoCode = this.kanjiInfoLabelList[i];
+        const j = entry.misc[kanjiInfoCode];
         result.push(
           this.kanjiInfoLabelList[i + 1].replace('&amp;', '&') +
             '\t' +
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -979,14 +979,15 @@ class RcxDict {
         if (!entryMatch) {
           continue;
         }
+        const [_, word, pronunciation, definitions] = entryMatch;
 
-        if (entryMatch[2]) {
-          result.push(entryMatch[1] + '\t' + entryMatch[2]);
+        if (pronunciation) {
+          result.push(word + '\t' + pronunciation);
         } else {
-          result.push(entryMatch[1]);
+          result.push(word);
         }
 
-        const t = entryMatch[3].replace(/\//g, '; ');
+        const t = definitions.replace(/\//g, '; ');
         result.push('\t' + t + '\n');
       }
     }
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -940,7 +940,6 @@ class RcxDict {
   }
 
   makeText(entry: DictEntryData | null, max: number): string {
-    let j;
     let t;
 
     if (entry === null) {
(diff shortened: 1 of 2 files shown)


extension/data.ts
```diff
@@ -940,60 +940,57 @@ class RcxDict {
...
-    let i;
-    let j;
-    let t;
-
     if (entry === null) {
       return '';
     }
 
-    const b = [];
+    const result = [];
```
melink14, line 947: In the long run, this type of structured output should probably be done with a template (such that the final text is easier to see in the code).

For HTML, I would naturally assume `lit-html` would be a good candidate since I use it for the options page.

However, I'm not sure if it would work for generating non-html text (haven't tried). If not maybe something like normal template literals would improve things (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) but we'd need to do something special for the list inputs...

Not a request to change anything but something I thought of when seeing your improvements!

extension/data.ts
```diff
@@ -940,60 +940,57 @@ class RcxDict {
...
-          this.kanjiInfoLabelList[i + 1].replace('&amp;', '&') +
-            '\t' +
-            (j || '-') +
-            '\n'
+      for (let i = 0; i < this.kanjiInfoLabelList.length; i += 2) {
+        const kanjiInfoCode = this.kanjiInfoLabelList[i];
+        const kanjiInfoName = this.kanjiInfoLabelList[i + 1];
+        const kanjiInfo = entry.misc[kanjiInfoCode];
+        result.push(
+          kanjiInfoName.replace('&amp;', '&') + '\t' + (kanjiInfo || '-') + '\n'
```
melink14, line 968: Thinking about these hyphens. I wonder if it would be better to replace empty string with hyphens from the very beginning (when we read the data from the dictionary) so we don't need to check so often later.

This one could at least be moved to line 966 but that depends on how things look after later refactors.

extension/data.ts
```diff
@@ -940,60 +940,57 @@ class RcxDict {
...
     } else {
       if (max > entry.data.length) {
         max = entry.data.length;
       }
-      for (i = 0; i < max; ++i) {
-        e = entry.data[i].entry.match(/^(.+?)\s+(?:\[(.*?)\])?\s*\/(.+)\//);
-        if (!e) {
+      for (let i = 0; i < max; ++i) {
+        const entryMatch = entry.data[i].entry.match(
+          /^(.+?)\s+(?:\[(.*?)\])?\s*\/(.+)\//
```
melink14, line 977: this is ripe to be converted to named capture groups probably. Just an FYI of course!

extension/data.ts
```diff
@@ -940,60 +940,57 @@ class RcxDict {
...
+        );
+        if (!entryMatch) {
```
melink14, line 979: The codecov failure is because this branch isn't covered I believe. (the first branch at the top is also not covered but since you didn't change those lines you shouldn't be impacted).

Since that's just a limitation of the coverage algorithm, there's no need to solve it now if you don't want. But I'll need to override the check.

melink14 (approved):
Great work keeping the PR small. It looks good to me with just some FYI comments. After you read/resolve stuff I'll merge it directly!

I hope you're not blocked by the review process since you can keep working on top of these changes with stacked branches!

extension/data.ts
bryanjenningz: That's a good idea - it would make the regex more readable. We can do this in a later pull request to keep this pull request small.

extension/data.ts
bryanjenningz: I decided to add some tests to get full test coverage for `makeText` so this problem is resolved now.

extension/data.ts
bryanjenningz: This is a good idea. I decided to do this with both `kanjiInfo` and `kanjiInfoName`.

extension/data.ts
bryanjenningz: Yeah, that might make the code feel more declarative and easier to follow. I'll keep that in mind in the future.

extension/test/data_test.ts
```diff
@@ -134,6 +140,48 @@ describe('data.ts', function () {
           '<word-1>\t<pronunciation-1>\t<definition-1>; <definition-2>; <definition-3>\n<word-2>\t<definition-1>; <definition-2>\n'
         );
       });
+
+      describe('when data entry is an invalid format', function () {
```
bryanjenningz, line 144: We might restructure the data types for the dict entries in the future so that the values are parsed and checked before this function is even called, so we might be able to delete these tests in the future.

I'll leave these tests in for now just so the current behavior is documented and so we won't have to deal with code coverage problems in `makeText` for now.

extension/test/data_test.ts
```diff
@@ -86,6 +86,12 @@ describe('data.ts', function () {
   });
 
   describe('makeText', function () {
+    it('returns an empty string if null dict entry is passed in', function () {
+      const emptyString = rcxDict.makeText(null, /* max= */ 1);
```
melink14, line 90: (super nit) It's a bit strange to call this `emptyString` already since that's what we're asserting later. Usually, this would just be something like `result` or `text`. By embedding the expecation in the actual value it might lead readers to not read the assertion line as closely leading to eventual drift.

I actually realize now this is similar to what you've done in the other tests and I wouldn't change it in this PR.

melink14 (approved):

extension/test/data_test.ts
bryanjenningz: Yeah, that makes sense. I can change all the variable names so they aren't like this.

PR #1625, opened 2023-06-17
refactor(makeText): Make kanji info labels an array of objects

bryanjenningz:
The goal of this pull request is to simplify the `makeText` function by changing the kanji info labels type definition.

Changes:
- Change `kanjiInfoLabelList` type to `{ code: string; kanjiInfoLabel: string }[]`
- Use a for-of loop to iterate over `kanjiInfoLabelList`
- Replace `&amp;` with `&` so that code in `makeText` doesn't need to have code that does that
  - `kanjiInfoLabelList` was [only used in `makeText`](https://github.com/search?q=repo%3Amelink14%2Frikaikun%20kanjiInfoLabelList&type=code) so this change is safe to make
- Rename `kanjiInfoLabelList` to `KANJI_INFO_LABELS`
- Make `KANJI_INFO_LABELS` private and static and readonly

This pull request doesn't change the behavior of `makeText` or the copy-to-clipboard feature - it only refactors `makeText` to be more simple.

### DIFF
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -601,7 +601,7 @@ class RcxDict {
     return entry;
   }
 
-  kanjiInfoLabelList: string[] = [
+  private kanjiInfoLabelList: string[] = [
     /*
         'C',   'Classical Radical',
         'DR',  'Father Joseph De Roo Index',
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -601,7 +601,7 @@ class RcxDict {
     return entry;
   }
 
-  private kanjiInfoLabelList = [
+  private KANJI_INFO_LABELS = [
     /*
         'C',   'Classical Radical',
         'DR',  'Father Joseph De Roo Index',
@@ -947,7 +947,7 @@ class RcxDict {
         result.push('\u90E8\u9996\u540D\t' + entry.bushumei + '\n');
       }
 
-      for (const { code, kanjiInfoLabel } of this.kanjiInfoLabelList) {
+      for (const { code, kanjiInfoLabel } of this.KANJI_INFO_LABELS) {
         const kanjiInfo = entry.misc[code] || '-';
         result.push(kanjiInfoLabel + '\t' + kanjiInfo + '\n');
       }
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -601,7 +601,7 @@ class RcxDict {
     return entry;
   }
 
-  private kanjiInfoLabelList: string[] = [
+  private kanjiInfoLabelList = [
     /*
         'C',   'Classical Radical',
         'DR',  'Father Joseph De Roo Index',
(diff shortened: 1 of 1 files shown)


extension/data.ts
```diff
@@ -613,32 +616,19 @@ class RcxDict {
         'K',  'Gakken Kanji Dictionary Index',
         'W',  'Korean Reading',
     */
```
bryanjenningz, line 618: I'm not sure if we want to keep this comment up here.

extension/data.ts
```diff
@@ -613,32 +616,19 @@ class RcxDict {
...
+    { code: 'L', kanjiInfoLabel: 'Heisig 5th Edition' },
+    { code: 'DN', kanjiInfoLabel: 'Heisig 6th Edition' },
+    { code: 'E', kanjiInfoLabel: 'Henshall' },
+    { code: 'DK', kanjiInfoLabel: 'Kanji Learners Dictionary' },
+    { code: 'DL', kanjiInfoLabel: 'Kanji Learners Dictionary 2nd Edition' },
+    { code: 'N', kanjiInfoLabel: 'Nelson' },
+    { code: 'V', kanjiInfoLabel: 'New Nelson' },
+    { code: 'Y', kanjiInfoLabel: 'PinYin' },
+    { code: 'P', kanjiInfoLabel: 'Skip Pattern' },
+    { code: 'IN', kanjiInfoLabel: 'Tuttle Kanji & Kana' },
```
bryanjenningz, line 629: I changed this `&amp;` to `&` since the `&amp;` was not used anywhere, so it was safe to make this change.

extension/data.ts
melink14: I guess this comment is meant to describe kanji info which is present in the dictionary but that we don't extract for rikaikun.

I'm not sure why we don't just include all of them (since they're optional) but having the comment here seems like a quick way to answer questions about missing indicies. (that said, this is probably out of date so maybe linking to the [canonical source of truth](http://www.edrdg.org/wiki/index.php/KANJIDIC_Project) would also be good)

That raises some additional questions to me though:
1. Should we periodically look at the fields available and add them to rikaikun (it is pretty easy to whitelist a new info field iirc)

2. Is replacing the comment with just a link to the canonical source enough. One thing this comment does is make it easy to tell which fields are intentionally not included but maybe it would be better to strive to include all fields and periodically check for new fields (I'm guessing update-db could be made to print a message when such an update is available)

tl;dr
Probably leave for now maybe adding the link to the documentation; we can think about the deeper questions later.

melink14 (approved):

extension/data.ts
bryanjenningz: I added a small explanation and the link you mentioned to the comment. Thanks!

melink14 (approved):

PR #1626, opened 2023-06-18
refactor(makeText): Use parseWordDictEntry in makeText

bryanjenningz:
The goal of this pull request is to replace inline word dict entry parsing with a call to parseWordDictEntry.

We can use parseWordDictEntry in other places in subsequent pull requests.

This pull request doesn't change any of the existing behavior - it only refactors the code.

### DIFF
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -601,7 +601,7 @@ class RcxDict {
     return entry;
   }
 
-  kanjiInfoLabelList: string[] = [
+  private kanjiInfoLabelList: string[] = [
     /*
         'C',   'Classical Radical',
         'DR',  'Father Joseph De Roo Index',
--- a/extension/data.ts
+++ b/extension/data.ts
@@ -94,7 +94,7 @@ type WordDictEntry = {
   definitions: string[];
 };
 
-export const parseWordDictFields = (line: string): WordDictEntry | null => {
+export const parseWordDictEntry = (line: string): WordDictEntry | null => {
   const fields = line.match(WORD_DICT_ENTRY_REGEX);
   if (!fields) {
     return null;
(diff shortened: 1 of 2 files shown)


bryanjenningz:
I stacked this branch on top of a previous branch so it contains the commits from the previous branch. The reason why these commits from the previous branch show up is because the previous branch's commits were squashed when the branch was merged.

bryanjenningz:
I'm going to close this pull request and figure out how to fix this problem where extra commits are showing up.

melink14:
> I'm going to close this pull request and figure out how to fix this problem where extra commits are showing up.

I saw you opened a new PR already but as a follow up what approach did you take to fix that problem?

In general, I think rebasing this PR's branch onto main after syncing should work, right?

bryanjenningz:
> In general, I think rebasing this PR's branch onto main after syncing should work, right?

Yes, it works if you rebase.

More specifically, doing this worked (it got rid of all the extra commits from the previous branch):

```bash
git checkout main
git pull upstream main
git checkout refactor-data.ts/part-1
git checkout -b test-branch
git rebase main
# Then you have to fix a series of merge conflicts, then `git add .` and `git rebase --continue`
git push origin HEAD
```

> I saw you opened a new PR already but as a follow up what approach did you take to fix that problem?

I ended up just copy/pasting the code to a new branch because it was faster, but rebasing using the commands above also worked when I tried it.

melink14:
I don't know the exact state of your branch but I think to avoid so many merge conflicts you can use `git rebase --onto` syntax instead of the default `git rebase` it allows you essentially tell git where it should start the rebase so you don't have to worry about merge conflict between the many commits and the single squashed commits.

I think this stackoverflow has a good explanation: https://stackoverflow.com/a/29916361

PR #1627, opened 2023-06-18
refactor(makeText): Create `parseWordDictEntry` function and use in makeText

bryanjenningz:
`parseWordDictEntry` will be used in `makeHtml` and the update-db.ts file in a subsequent pull request. Unifying the parsing logic to a single function will reduce repetition and make the code easier to understand and easier to test and change.

Changes to existing behavior:
- If there's a malformed dict line that gets looked up, it gets logged. There is currently only 1 line in the word dictionary that is malformed (which is `４° [しど] /`) so it's not a big change to existing behavior.


extension/data.ts
```diff
@@ -86,6 +86,27 @@ interface DeinflectionRuleGroup {
   rules: DeinflectionRule[];
 }
 
+const WORD_DICT_ENTRY_REGEX = /^(.+?)\s+(?:\[(.*?)\])?\s*\/(.+)\//;
```
bryanjenningz, line 89: I plan on changing this regex to be more readable with named capturing groups, but for now I'm keeping it the same as it was before to reduce the amount of change in this pull request.

extension/data.ts
```diff
@@ -86,6 +86,27 @@ interface DeinflectionRuleGroup {
...
+
+type WordDictEntry = {
+  word: string;
+  pronunciation: string | null;
```
melink14, line 93: In the interest of avoiding null in types I wonder if it's possible to just make this `string` type and use empty string if there's no pronunciation? That tends to make things easier to reason about and allows avoiding guards.

Of course, if we always need to check for empty string instead, then there's not much difference in which case keeping it like this is maybe more consistent.

extension/data.ts
```diff
@@ -86,6 +86,27 @@ interface DeinflectionRuleGroup {
...
+  definitions: string[];
+};
+
+export const parseWordDictEntry = (line: string): WordDictEntry | null => {
+  const [, word, pronunciation, definitions] =
+    line.match(WORD_DICT_ENTRY_REGEX) || [];
+  if (!word || !definitions) {
```
melink14, line 100: Is there any reason we would expect this to happen normally? It seems like it can only happen in an error case or new type of dictionary entry. maybe logging a warning would be good if it's unexpected...

extension/data.ts
```diff
@@ -86,6 +86,27 @@ interface DeinflectionRuleGroup {
...
 
+const WORD_DICT_ENTRY_REGEX = /^(.+?)\s+(?:\[(.*?)\])?\s*\/(.+)\//;
+
+type WordDictEntry = {
+  word: string;
+  pronunciation: string | null;
+  definitions: string[];
+};
+
+export const parseWordDictEntry = (line: string): WordDictEntry | null => {
```
melink14, line 97: In this case, we're exporting this function soley in order to test it but in the end increases it visibility for any other use case as well.

In order to avoid API creep, a good best practice is to have all methods/variables as non-accessible as possible for the needs of the software.

In this case, is there any reason this method can't be private inside of RcxDict? Isn't all of this logic already tested via the public APIs of `makeText`?

If we did want a generic entry parsing function, I'd recommend putting it in its own module so that it can be treated and maintained as an independent utility.

-- Whoops! I just remembered that you did say you planned to use it in update-db.ts as well so in that case I would go with the last recommendation and extract it to it's own module. This will keep files small and make them very easy to reason about; often called the Single Responsibilty Principle.

extension/test/data_test.ts
```diff
@@ -85,6 +85,67 @@ describe('data.ts', function () {
     });
   });
 
+  describe('parseWordDictEntry', function () {
+    describe('with a valid word dict line', function () {
+      it('parses a word dict entry from a word dict line which has a word, pronunciation, and definitions', function () {
```
melink14, line 90: In these test names, we're saying that the correct outcome is that the method is going to "parse a word dict entry" but it's that leaves some ambiguity on what correct parsing is.

In this case, the key to parsing is mapping the word to the `word` property and so forth. It's a bit verbose to do it all at once so I actually recommend testing each output property in it's own test like: 'parses the word part of the entry and returns in word property'. (then you would have three similar tests with different assertions basically)

This has several benefits:
- Failures are more precise since the exact piece of extraction is called out.
- Writer intention is extremely clear so as documentation the test is improved.
- Future additions to entries can be tested primarily in new tests; it's an anti-pattern to update pre-existing tests for new functionality since that reduces their value as regression finders.

The down side is additional tests but we're not close to any size constraints on a project of this size! Writing them is very easy since the structure is the same after the first one.

Let me know what you think!

extension/test/data_test.ts
```diff
@@ -85,6 +85,67 @@ describe('data.ts', function () {
...
+    describe('with an invalid word dict line', function () {
+      it('returns null when you pass in an empty word dict line', function () {
+        const emptyWordDictLine = '';
+
+        const wordDictEntry = parseWordDictEntry(emptyWordDictLine);
+
+        expect(wordDictEntry).to.equal(null);
+      });
+
+      it('returns null when you pass in a word dict line with badly formatted definitions', function () {
```
melink14, line 128: I wonder if it's worth being explicit about 'badly formatted definitions'.  I guess it's having a separator but empty defintions? If it's hard to think of a good descriptor this isn't bad but I thought of it while reading.

melink14 (changes requested):
Looks good; the main comments are around moving it to it's own module as well as some suggestions around test structure. Let me know your thoughts!

extension/data.ts
bryanjenningz: It only happens if there's an improperly formatted dictionary line. Right now there's only a single line in the dictionary that's improperly formatted: `４° [しど] /`.

Yeah logging makes sense.

If we use this parsing function in update-db.ts in the future then these logs will show up and so we'll know if there are any additional improperly dictionary lines that get added when the dictionary changes in the future.

extension/data.ts
bryanjenningz: Maybe the only benefit to using `null` here would be that it forces you to explicitly set a default value or explicitly handle empty pronunciation values. This isn't a huge benefit. I could also see that also as a negative depending on the situation - it might be annoying to have to explicitly set a default value, so I'm fine with just setting this as an empty string instead of using `null`.

extension/test/data_test.ts
bryanjenningz: I tried this out and it seems reasonable. The tests are more specific now and it's easier to see what specifically is failing when a test fails. Thanks!

extension/data.ts
bryanjenningz: I put this in its own file so it can be used in update-db.ts later.

extension/test/data_test.ts
bryanjenningz: I reworded this. Thanks!

extension/parse-word-dict-entry.ts
```diff
@@ -0,0 +1,21 @@
...
+  word: string;
+  pronunciation: string;
+  definitions: string[];
+};
+
+export const parseWordDictEntry = (line: string): WordDictEntry | null => {
+  const [, word, pronunciation, definitions] =
+    line.match(WORD_DICT_ENTRY_REGEX) || [];
+  if (!word || !definitions) {
+    console.log('Improperly formatted word dict line', line);
```
bryanjenningz, line 13: The downside of having a console.log here is it shows up in the test output. I'm wondering if I should silence it somehow for tests.

melink14:
Is there a problem for it to show up in the test output? In fact, adding a
new test for the log would make sense to me since its an external side
effect

I assume it only shows up only in the tests which are triggering the
malformed branch?

2023年6月20日(火) 8:21 Bryan Jennings ***@***.***>:

extension/parse-word-dict-entry.ts
melink14: Is there a problem for it to show up in the test output? In fact, adding a
new test for the log would make sense to me since its an external side
effect

I assume it only shows up only in the tests which are triggering the
malformed branch?

bryanjenningz: I guess it just looks kind of weird when I see that the tests all pass but the logs say "Improperly formatted word dict line". It makes me feel like something is wrong and the tests aren't catching it even though it's logging as expected. It seems like it clutters the logs a little bit, but it's not a big deal.

Yeah, it only shows up in tests which are triggering the malformed branch.

bryanjenningz: I added a test which checks that the console.log function is called for malformed dict lines.

melink14 (approved):

extension/test/parse-word-dict-entry_test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+use(sinonChai);
+
+describe('parseWordDictEntry', function () {
+  afterEach(function () {
+    sinon.restore();
+  });
+
+  describe('with a valid word dict line', function () {
+    describe('when word dict line has pronunciation', function () {
```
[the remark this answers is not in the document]
bryanjenningz: Yeah, the wording 'when word dict line has all parts' is a bit more clear. I agree.

melink14 (approved):

PR #1635, opened 2023-06-21
refactor(dict-entry): Separate kanji and word dict entry types

bryanjenningz:
This pull request separates the `DictEntryData` type into `KanjiDictEntryData` and `WordDictEntryData`, which makes it more clear what type of data a function is working with.

Now we can differentiate between the different dict entry types by checking the `type` property and we can see if it is a `'KANJI'` dict entry or a `'WORD'` dict entry.

This pull request is a refactor - it doesn't change any of the existing behavior.


bryanjenningz:
I think it makes more sense to do this later, so I'm going to close this pull request for now.

PR #1974, opened 2024-02-14
fix(deinflect):  Only deinflect って and friends to く for 行く

### LINKED ISSUE #1920: Rikaikun thinks って can be an inflection of く verbs
For example, やって will give results for various やく verbs; this is probably not too commonly seen since if a kanji is used, it will never appear using the wrong inflection.

This is probably because an inflection was added for 行く which has an irregular conjugation to 行って and 行った.
https://github.com/melink14/rikaikun/blob/b908eb0f92070bb322697f12f6e4f53fc767f780/extension/data/deinflect.dat#L263

Instead, we should add the irregular conjugations as full lines like we do for 来る and する. i.e.:

いって　いく

I don't think we need to update the type mask (explained in this post: https://github.com/melink14/rikaikun/pull/988#issuecomment-1107340408) since it still operates as a ru verb and isn't as irregular as suru/kuru.

Edit:

Found some other similar lines like:
https://github.com/melink14/rikaikun/blob/b908eb0f92070bb322697f12f6e4f53fc767f780/extension/data/deinflect.dat#L94

Which are different versions of って with auxillary verbs. Probably should check for all く lines and fix any that have small て in the first part.

bruno-queiroz:
We don't match the kana form in order to not incorrectly match いって to 往く, which would be incorrect. 行く is almost always written with kanji so it seems like the right tradeoff.
 
Fixes #1920


extension/data/deinflect.dat
```diff
@@ -363,3 +360,6 @@ past negative
 でございましょう	でございます	8320	18
 でございまして	でございます	8320	21
 であれば	でございます	8320	17
+いって	いく	384	0
```
melink14, line 363: Ah, sorry for the misunderstanding but the index isn't into the list of similar inflections but into the reason list at the top of the file. That means, we should keep the final index numbers the same since the conjugation reasons are the same.

(14, 21, 3)

extension/data/deinflect.dat
```diff
@@ -363,3 +360,6 @@ past negative
...
+いった	いく	384	1
```
melink14, line 364: The 384 here might be wrong. It looks like it corresponds to a hex value of 0x0180 which means a deinflection of OTHER to RU VERB (See https://gist.github.com/lynn/07f7ce3c314223d2aca19ec2bb0540cd). But in this case iku is more like a U VERB than a RU VERB right?

In that case it seems like we should use 0x0280 which is 640 decimal.

In that way it looks like we probably can reuse the values from the previous lines here as well.

This is also could have been why 行って stopped working. Since it wouldn't have recognized it as a RU verb.

melink14 (changes requested):
Left two comments which should answer the questions you had in the PR description. Sorry if my previous explanations were unclear. (Also sorry that this part of the code base hasn't been cleaned up to be easier to edit)

extension/data/deinflect.dat
bruno-queiroz: I did those changes but apparently the 行って problem still persists for some reason.

bruno-queiroz:
> Left two comments which should answer the questions you had in the PR description. Sorry if my previous explanations were unclear. (Also sorry that this part of the code base hasn't been cleaned up to be easier to edit)

no problems, thanks for explaining and helping me.

extension/data/deinflect.dat
melink14: Nothing is obviously wrong so I guess the next step is to step through the relevant lookup code in data.ts with devtools and figure out why it's not matching. I'll try to find some time today or tomorrow to checkout your PR and debug, though feel free to try yourself. :)

melink14: Alright, I debugged it a bit and there were two problems. 

1. I forgot how simple the logic for this is. It doesn't actually convert kanji to kana for the deinflection.  So we need to actually add the kanji to the file like:
```
行って	行く	640	21
行った	行く	640	14
行っちゃう	行く	514	3
```

This works well but means that if someone is looking up the kana form it won't match. We could add the kana form as well but then words like 往く will also match. I feel like avoiding wrong results is better than not having every corner case  covered since 行く is usually written with the kanji.

2. There was a small typo in the いっちゃう entry where you used a space instead of a tab. I only saw it after I tested the above fix and it still didn't work. ^_^;

bruno-queiroz: Now looks like it's working, thank you for helping me and sorry for bothering you with a lot of questions. This was actually my first time doing open source.

bruno-queiroz:
Why did the tests failed this time? 🤔

melink14:
> Why did the tests failed this time? 🤔

Not sure; for now I'm just re-running them in case it was a fluke.

melink14:
> > Why did the tests failed this time? 🤔
> 
> Not sure; for now I'm just re-running them in case it was a fluke.

Looks like that worked. I will lightly edit the PR title/description for style and then approve which will start the merge process. (The PR title and body become the commit message so they're a bit more important than in other repos)

melink14 (approved):

melink14:
@all-contributors please add @bruno-queiroz for code.

bruno-queiroz:
Thank you 😁

PR #2158, opened 2024-08-18
refactor: Remove non-null assertion from inlineDisable

### LINKED ISSUE #2154: Remove nonnull assertion operator from `inlineDisable`
https://github.com/melink14/rikaikun/blob/8fd4bf403742fd6041767c79d87f5e080aa6d033/extension/rikaichan.ts#L217-L222

This nonnull assertion operator seems unnecessary due to the guard above, but it seems like indexing into certain arrays means typescript is not sure something hasn't changed.

The workaround is to save the value to a variable before checking it. (ref https://stackoverflow.com/questions/75640224/type-narrowing-is-not-working-in-for-loop)

Interestingly this is already done for `tabs` in the previous block. One way to get a variable for free is to switch to a `for..of` loop which always gives each iteration a variable. I don' tthink we're using the index for anything else so that is probably the cleanest solution.

ChinoUkaegbu:
Fixes #2154 
Saved the value to a variable (tabId) before checking explicitly if it was undefined and changed loop structure for cleaner code

### DIFF
--- a/extension/rikaichan.ts
+++ b/extension/rikaichan.ts
@@ -209,16 +209,17 @@ class RcxMain {
 
     // Send a disable message to all browsers
     chrome.windows.getAll({ populate: true }, (windows) => {
-      for (let i = 0; i < windows.length; ++i) {
-        const tabs = windows[i].tabs;
+      for (const window of windows) {
+        const tabs = window.tabs;
         if (tabs === undefined) {
           continue;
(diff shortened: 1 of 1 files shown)


melink14 (approved):
You got flagged for tests but that's because the previous lines were already untested. I'll override the check in this case. Thanks!

PR #2176, opened 2024-08-26
fix(google-docs): Set docs_annotate_canvas property in a MV3 compliant manner

### LINKED ISSUE #2175: Not popup for google doc on microsoft edge
**Describe the bug**
Not popup for google doc on microsoft edge 

**To Reproduce**
Install and use on Edge

**Expected behavior**
Popup on Edge contents

**Screenshots**

**System details (OS, browser version, rikaikun version)**

- OS: MacOS Sonoma 14.6
- Browser Version: Version 128.0.2739.42
- rikaikun Version: 3.0.0

**Additional context**

melink14:
New CSP changes means we can no longer inject a script into the MAIN world from the ISOLATED world.

Instead, we use the new `world` key to inject a simple script directly into the MAIN world.

One caveat of this is that we can no longer read the extension ID dynamically (as it's not available in MAIN world) so instead we hardcode it into the script as it extremely stable.

Fixes #2175


melink14:
Code coverage dropped because total lines of code dropped. ignoring.

PR #2323, opened 2024-12-17
build(bundler): Bundle using vitejs instead of snowpack

### LINKED ISSUE #937: Migrate away from snowpack
snowpack is in maintenance mode and new plugins and dependency releases are probably not expected.
Already, I use an override to force a newer esbuild but more recently I also was troubled by the lack of inline CSS support.

So far I can get the same basic natural support with esbuild + esinstall and have a branch with proof of concept.

https://github.com/aheissenberger/esbuild-plugin-text-replace is a potential plugin (I think I already started using) to replace imports with pointers to esinstall's package.

melink14:
This is essentially a one to one replacement to snowpack trying to keep things as unbundled as possible. Some notable steps:

- We unbundle by using a glob to mark all app files as inputs, this is preferred to using `preserveModules` since it's nice to still bundle node_module deps.
- To make node_module deps a bit more tractable, we use `manualChunks` to group them by top level package.
- Debugging in vscode required a webRoot in launch.json, but that might have been needed for snowpack given I haven't used it in a long time.
- Coverage lowered but that's because I was able to include just the SUT code while snowpack also included the test files which gave a big bump in coverage.
- Removed the override for esbuild that was required due to snowpack not being maintained.
- Needed to patch the vite wtr plugin so that it would forward headers. Since our css is inside our src directory, without 'text/css' header, during testing the css would be transformed into JS.
- Needed to patch vite-multiple-assets in order to fix a type error caused by duplicate interfaces.
- Needed to install rollup as a dev dependency to prevent a type failure from vite-multiple-assets.
- For some reason, needed to change tsconfig module resolution to 'bundler' that makes sense, but not sure why it stopped working with this change.
- Needed to define `global` in the WTR test HTML as vite doesn't provide it by default.
- File extensions are not transformed to js during serving so any references to code for tests should use ts suffix (chrome_stubs.ts in this case)
- To make paths consistent, I needed to set WTR root dir to 'extension' to match the vite root dir.

For more test commentary see https://github.com/remcovaes/web-test-runner-vite-plugin/issues/10.

Later I'll create bugs to remove the various patches I've added based on the upstream bugs.

Fixes #937


melink14:
Will merge, ignoring codecove/project since the decrease is caused by test files not being considered anymore, the new number is the accurate one.

melink14:
@Mergifyio refresh

PR #2324, opened 2024-12-18
refactor: Remove unnecessary lint disable

melink14:
It seems some bug was fixed in the background, since if it was a dependency change it would have caused a failure then.


melink14:
Merging over known breakage to make fixing more clean.

PR #2553, opened 2025-05-11
build(eslint): Remove eslint disable comments for broken no-setup-in-description rule

melink14:


melink14:
Just a postscript that I opened a bug for this at https://github.com/lo1tuma/eslint-plugin-mocha/issues/383 and we should revert this CL when fixed.

PR #2662, opened 2025-07-16
feat: Add Z key to zoom popup (to ease reading small multi-stroke kanji)

akx:
This PR adds a new hotkey, <kbd>Z</kbd>, to toggle between a couple different zoom levels for the popup (lazily implemented with `transform: scale`).


melink14:
Thanks for the PR! I'll pull it and test locally later but in the mean time do you think you could add visual tests for each zoom level? (Basically, for each zoom level and to confirm it goes back to default)

It should be fundamentally the same as the something like https://github.com/melink14/rikaikun/blob/6b914377c45af643af7a4ef165c8e45b67a5228b/extension/test/e2e_visual_test.ts#L112

The main difference is that I don't think we can just do one color as this is behavior is color agnostic. (At least in my opinion, I'm open to other ideas though. :)

akx:
Sure. Added a simple test (that I couldn't run locally, but here's hoping CI takes care of it).

extension/test/e2e_visual_test.ts
```diff
@@ -133,6 +133,30 @@ describe('Visual Regression Tests', function () {
     });
   });
 
+  describe('zoom', function () {
+    it('should cycle through zoom levels', async function () {
```
melink14, line 137: Instead of just one test for all zoom levels can we have a single test for each one. Splitting the behaviors will make maintaining the tests a bit more easier in the futuer I think.

example: 'should render at zoom level 1 when 'z' pressed one time' and so forth. Then each test is basically the same except with more key presses. :)

extension/test/e2e_visual_test.ts
```diff
@@ -133,6 +133,30 @@ describe('Visual Regression Tests', function () {
...
+      const span = insertHtmlIntoDomAndReturnFirstTextNode(
+        '<span>虫眼鏡</span>'
+      ) as HTMLSpanElement;
+
+      await triggerMousemoveAtElementStart(span);
+      // Tick the clock forward to account for the popup delay.
+      clock.tick(150);
+      await waitForVisiblePopup();
+      // Cycle through zoom levels by pressing 'z'.
+      await takeSnapshot('word-entries-zoom-0');
```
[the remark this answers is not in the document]
akx: I guess I'll also remove the last snapshot then? So something like

```
      await waitForVisiblePopup();
      await sendKeys({ press: 'z' }); // Zoom in.
      await takeSnapshot('word-entries-zoom-1');
      await sendKeys({ press: 'z' }); // Zoom in more.
      await takeSnapshot('word-entries-zoom-2');
      await sendKeys({ press: 'z' }); // Zoom in even more.
      await takeSnapshot('word-entries-zoom-3');
      await sendKeys({ press: 'z' }); // Return to unzoomed state.
```
?

screenshots/Chromium/baseline/word-entries-zoom-1.png
melink14, line 1: I haven't had a chance to test it yet, but is this what it actually looks like or is the screenshot just not handling zoom well.

Right now it looks the same size but cropped. 😆

melink14 (changes requested):

extension/test/e2e_visual_test.ts
```diff
@@ -133,6 +133,29 @@ describe('Visual Regression Tests', function () {
...
   });
 
+  describe('zoom', function () {
+    for (const level of [1, 2, 3, 4]) {
+      it(`should work at zoom level ${level}`, async function () {
+        const clock = sinon.useFakeTimers();
+        const span = insertHtmlIntoDomAndReturnFirstTextNode(
+          '<span>虫眼鏡</span>'
+        ) as HTMLSpanElement;
+        await triggerMousemoveAtElementStart(span);
```
melink14, line 143: Super nit, but can you add a newline before this line and after line 149 to separate the 'parts' of the test similar to the other tests. (It's a small boon to see arrange, act, and assert sections independently)

melink14 (changes requested):
Looks good modulo the final nits. I'll finish local e2e testing over the weekend so it's ready for next weekly push.

Thanks again.

